PR #23, opened 2022-06-07
Fix typo preventing the creation of a new workspace

thomasp-ms:
Just remove 2 commas that had no business being there and prevented creating new workspaces


dkmiller (approved):
Thanks for the fix!

ShizeSu06 (approved):

PR #53, opened 2022-09-08
Small improvements

thomasp-ms:
- align compute names in the experiment with what's created by the auto-provisioning
- add a reminder in the README that people will need to log in


garg-amit (approved):
Looks good to me!

PR #55, opened 2022-09-08
To store aggregated models in a single folder

garg-amit:
## Purpose

* Aggregated models are to be stored in a single folder on the datastore (path: 
 "< date>_<random_num>/round_<round_num>/model.pt")

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[x] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

Submit a pipeline and go to the storage account to see if the aggregated models are stored in a single folder.


.github/workflows/python.yaml
```diff
@@ -7,7 +7,7 @@ on:
   pull_request:
     branches:
     - main
-    - release-sdkv2-mvp
+    - release-sdkv2-iteration-02
```
thomasp-ms, line 10: good catch!

examples/pipelines/fl_cross_silo_basic/submit.py
```diff
@@ -123,15 +126,30 @@ def custom_fl_data_path(datastore_name, output_name, unique_id="${{name}}", roun
...
 
+def getUniqueIdentifier(length):
+    """Generates a random string and concatenates it with today's date
+
+    Args:
+        length (int): length of the random string
+
+    """
+    str = string.ascii_lowercase
+    date = datetime.date.today().strftime("%B_%d_%Y_")
```
thomasp-ms, line 144: I suggest YYYY-MM-DD instead this format, as it will be easier to identify the results of a given run

examples/pipelines/fl_cross_silo_basic/submit.py
```diff
@@ -123,15 +126,30 @@ def custom_fl_data_path(datastore_name, output_name, unique_id="${{name}}", roun
...
+    return date + "".join(random.choice(str) for i in range(length))
+
+
+pipeline_identifier = getUniqueIdentifier(8)
```
thomasp-ms, line 148: Let's not have the length hidden and hard-coded like this. At the minimum let's put it in a "global" variable, but a better solution would be to treat this as an optional parameter with a default value of 8.

examples/pipelines/fl_cross_silo_basic/submit.py
```diff
@@ -123,15 +126,30 @@ def custom_fl_data_path(datastore_name, output_name, unique_id="${{name}}", roun
...
+
+
 @pipeline(
-    description="FL cross-silo basic pipeline",
+    description=f'FL cross-silo basic pipeline and the unique identifier is "{pipeline_identifier}" that can help you to track files in the storage account.',
```
thomasp-ms, line 152: I like how you added this to the description

thomasp-ms (changes requested):

examples/pipelines/fl_cross_silo_basic/submit.py
garg-amit: Ok. Done

examples/pipelines/fl_cross_silo_basic/submit.py
garg-amit: Agree. Provided a default value.

thomasp-ms (approved):

PR #56, opened 2022-09-13
Factored code for submitting FL cross-silo basic

jfomhover:
This pull requests proposes another way of formatting the FL submission code. It draws from the lessons of the shrike FL API, while attempting to get rid of the baggage dependencies shrike adds to developer code.

Here's the main characteristics of this proposal:
- all configuration management is left for user to do as they please (no core dependency on hydra, just a suggestion)
- writing an FL pipeline consists in coding 4 python fonctions only (no python object required)
- the contract between those 4 functions is flexible, but written explicitely in a clean pythonic way
- the contract is enforced and verified by the factory (testing if each method returns the right output, providing user readable exceptions whenever error happen)
- all the runsettings are taken care of by factory.

The code also provides a "soft validation" mechanism to enforce affinity checks between compute and datastores.

To get the quick understanding of the proposal, go directly to the `submit.py` script.

TODO:
- [x] add a --force argument to make sure users understand risk before submitting


examples/pipelines/fl_cross_silo_factory/submit.py
```diff
@@ -0,0 +1,308 @@
...
+
+try:
+    credential = DefaultAzureCredential()
+    # Check if given credential can get token successfully.
+    credential.get_token("https://management.azure.com/.default")
+except Exception as ex:
+    # Fall back to InteractiveBrowserCredential in case DefaultAzureCredential not work
+    credential = InteractiveBrowserCredential()
+
+# Get a handle to workspace
```
ShizeSu06, line 77: I'm a bit confused about why we try connect to workspace using config.json first, then try config.yaml if that not working.

Suggestion: In my opinion, better to always use the aml workspace in config.yaml for the FL experiment (since that file is the FL experiment config yaml file and thus should be the targeted aml ws). Adding the case of loading ws from config.json file in this submission script might just add confusion/complexity rather than providing clarity/help, and I suggest to remove it from this submission script.

examples/pipelines/fl_cross_silo_factory/fl_factory.py
```diff
@@ -0,0 +1,503 @@
...
+        self,
+        silo_preprocessing,
+        silo_training,
+        orchestrator_aggregation,
+        iterations=1,
+        **training_kwargs,
+    ):
+        @pipeline(
+            description=f'FL cross-silo basic pipeline and the unique identifier is "{self.unique_identifier}" that can help you to track files in the storage account.',
+        )
```
ShizeSu06, line 119: preprocessing step at silo better to be optional in fl-factory logics, and fl-factory logics need to handle the scenario whether or not users provide/use "preprocessing component" step. Refer to details for https://github.com/Azure/shrike/blob/main/shrike/pipeline/federated_learning.py, where preprocessing and post-processing step are optional, only silo-train and server-agg are hard requirement for FL API.

It is common for scenarios where customer silos don't need any extra preprocessing on the silo's data (e.g., already cleaned, or the silo's data needs offline heterogeneous preprocessing, e.g., different languages data from different companies, and can't use a shared common preprocessing step here.)

examples/pipelines/fl_cross_silo_factory/fl_factory.py
```diff
@@ -0,0 +1,503 @@
+"""Factory to build a basic FL pipeline."""
```
ShizeSu06, line 1: Would be good to have a very clear instruction (can be a separate readme about using such fl_factory to write/submit fl_pipeline experiment) about: whether users need to touch fl_factory.py, and if yes, when and how? Ideally, modify the config file would be enough for users, and touch fl_factory.py API code is not needed (except for some very rare case).

examples/pipelines/fl_cross_silo_factory/submit.py
```diff
@@ -0,0 +1,308 @@
+"""Federated Learning Cross-Silo pipeline built by a factory class.
```
ShizeSu06, line 1: It would be good to have very clear guidance (either at top of this demo submit.py file, or in a separate readme who also explained how to use fl_factory to write/submit fl experiment.) on where users need to touch/modify in such submission script when adapting it to their own FL scenario. E.g., based on my reading, roughly speaking, seems users might need to touch config.yaml+ section C (FL steps) + section D (adjust accordingly to their FL config)?

examples/pipelines/fl_cross_silo_factory/fl_factory.py
jfomhover: We could make each of those steps an optional argument in the builder, I'll refactor in this direction.

examples/pipelines/fl_cross_silo_factory/fl_factory.py
jfomhover: For now I'd suggest do not touch factory at all, but I will add a disclaimer in the header to explain why.

examples/pipelines/fl_cross_silo_factory/submit.py
jfomhover: Absolutely, let me write some more guidance in the beginning of the file. We will need a proper tutorial on this but I'd like to take more time to write it up in a distinct PR.

examples/pipelines/fl_cross_silo_factory/submit.py
jfomhover: This part is common between the factory submit and the native submit. The discussion here could be between 3 ways to provide aml references:
- provide as command line arguments to the submit.py script (ex: `--workspace NAME`)
- provide in yaml
- provide in config.json

Maybe a good order of precedence could be:
- first check cli arguments,
- in no cli, check yaml
- if no yaml, use config

Since this is a common piece of code between native and factory, I propose to file this as an open issue for this september release and figure it out for both scenarios.

jfomhover: Filed as #79

examples/pipelines/fl_cross_silo_factory/submit.py
jfomhover: Filed as #80

examples/pipelines/fl_cross_silo_factory/submit.py
```diff
@@ -0,0 +1,320 @@
...
+
+# 3. use a pipeline factory method
+
+pipeline_job = builder.build_basic_fl_pipeline(
+    # building requires all 3 functions provided as argument below
+    silo_preprocessing,
+    silo_training,
+    orchestrator_aggregation,
+    # RESERVED: this kwarg is for building iterations
+    iterations=YAML_CONFIG.training_parameters.num_rounds,
```
ShizeSu06, line 289: to be consistent and more clearer to users, would that be better to also use "iterations" rather than "num_rounds" in the FL config.yaml file? That is also what we did in shrike FL API, i.e., we always use the term "iteration".

ShizeSu06 (approved):

PR #57, opened 2022-09-16
Secure internal silos

thomasp-ms:
## Purpose
* This PR introduces a PowerShell script to secure a vanilla FL setup that was created with the `ProvisionVanillaSetup.ps1` script.
* By "secure a vanilla setup" I mean the following:
  * create and assign to the orchestrator compute a managed identity;
  * for each silo, do the following:
    * create and assign to the silo compute a managed identity;
    * create a datastore for reading the silo input data from a pre-existing storage account;
    * create data assets pointing to the training and test data in the pre-existing storage account;
    * create a "sharing" storage account to safely communicate model weights between the silo and the orchestrator;
    * create a "sharing" datastore associated to the "sharing" storage account;
    * give the orchestrator's and silo's identities Write permissions to the "sharing" storage account.
* The script does NOT give the silos' identities read access to the pre-existing storage account containing the highly sensitive silo data. I made this decision because in practice, I'm not sure whether the users will have the permissions to do that. They will likely need to reach out to someone for that one-off action.
* To successfully use the script, one must have _Contributor_ access to the subscription (for creating all the resources), and _Owner_ access to the workspace resource group (this is to allow for adjusting permission to the "sharing" storage accounts, which are all created under the workspace resource group).

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
* First, adjust a few values in `setup_info.yml`:
  * use the SubscriptionID value for the "Vienna PM 2" subscription and use `thopo-vanilla-fl-2-rg` as a resource group (unless you already have Owner role in a subscription or resource group, in which case, you can use those).
  * feel free to change the silos' names, but _do not change the `storage_account` and `container` values  (since they must be pointing to some pre-existing storage); do not change the silos locations either (they are actually not used for the computes right now due to a bug with a CLI, but eventually this will be fixed and we will want to create the silos computes in the same regions as the pre-existing storage accounts).
* Start the docker image as explained [here](https://github.com/Azure-Samples/azure-ml-federated-learning/blob/main/mlops/internal_silos/README.md).
* If it hasn't been done already, run the `ProvisionVanillaSetup.ps1` script to provision a vanilla setup.
* Run the `SecureVanillaSetup.ps1 script`.
  * Unless you have Owner access to the workspace resource group, you won't be able to change the permissions for the "sharing" storage accounts. This is expected, and will be clearly documented in the soon-to-come doc PR. Meanwhile, just reach out to me and I will grant access.
* Reach out to me so I can give the silos computes access to the pre-existing storage accounts.  

## What to Check
* Check that the example job can be submitted (this will require modifying the output datastores for every component).
* Try to introduce a mistake by having silo _n_ read from/write to the datastore associated with silo _m_. The job should fail.

## Future work
* Provide documentation for this step. 
* Adjust example job to use the newly introduced datastores.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -130,3 +130,6 @@ dmypy.json
 
 # if you use an AzureML config file locally
 config.json
+
+mlops/internal_silos/YAML/setup_info_local.yml
+examples/pipelines/test_job
--- a/mlops/internal_silos/YAML/setup_info.yml
+++ b/mlops/internal_silos/YAML/setup_info.yml
@@ -1,16 +1,22 @@
 subscription_id: <Your-SubscriptionId>
 workspace:
-  name: vanilla-fl 
+  name: thopo-vanilla-fl-2
   region: westus2
-  resource_group: vanilla-fl-rg
+  resource_group: thopo-vanilla-fl-2-rg
   orchestrator_compute_name: cpu-cluster
 silos:
   - silo_01:
     name: silo-wus2
     region: westus2
+    storage_account: thopostaccwus2test
+    container: mnist
   - silo_02:
     name: silo-eus2
     region: eastus2
+    storage_account: thopostacceus2test
+    container: mnist
   - silo_03:
     name: silo-weu
-    region: westeurope       
+    region: westeurope
+    storage_account: thopostaccweutest
+    container: mnist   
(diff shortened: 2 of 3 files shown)


thomasp-ms:
Abandoning this PR, since we're now doing the securitization via Bicep templates.

PR #58, opened 2022-09-16
Native bicep provisioning script for vanilla demo setup

jfomhover:
This PR creates a pure BICEP script for deploying all the resources for a vanilla demo with internal silos.

For a given set of regions, it will:
- provision an AzureML workspace and compute cluster for orchestration
- per region, provision a silo (1 storage with 1 private containers, 1 compute, 1 UAI)
- set the permissions using the UAI between silo compute and silo containers

The PR also aligns the configuration of the basic pipeline with this setup script naming conventions. The graph had issues submitting which are now fixed.

Here's a couple todos out of this PR:
- #61 : the current permissions model is not ideal, it relies on having each silo compute having R/W on the orchestrator.
- #60 : figure out in the instructions how to upload the data smoothly (current not great, but ok?)


examples/pipelines/fl_cross_silo_basic/config.yaml
```diff
@@ -9,22 +9,43 @@ aml:
...
-      datastore: workspaceblobstore
-      training_data_path: "azureml:mnist-train-03:1"
-      testing_data_path: "azureml:mnist-test-03:1"
+    - compute: cpu-cluster-eastus2
+      datastore: silo_datatore_eastus2_private
+      # in this demo, we're using public data from a url instead
+      training_data:
+        type: uri_file
+        mode: 'download'
+        path: https://azureopendatastorage.blob.core.windows.net/mnist/processed/train.csv
```
garg-amit, line 22: One reason for wrapping these train/test sets in different data assets for each silo was to have a feel in the AML UI that the data is not shared among them and each silo has its own dataset. I also agree, with the new approach - it is pretty straightforward to handle data. We might need to think about this (maybe in the next iteration).

garg-amit (approved):
Looks good to me!

examples/pipelines/fl_cross_silo_basic/config.yaml
jfomhover: Yes I think we need to provide a "feel" in our demo. I filed an issue #60 to figure out what's the best strategy here. I propose to merge this one and add a PR once we define the best approach.

PR #69, opened 2022-09-21
Remove redundant files from the mlops directory

garg-amit:
## Purpose

* Remove internal & external silos dir as the provisioning is taken care of by the bicep dir
* We can always revisit the GitHub history if we want to reuse a few components from the internal/external dir

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```


jfomhover (changes requested):
I had not noticed we had some binary tgz files of the dataset. Could you actually NOT remove them, but instead move those files under examples/data/mnist/ ?

We might need this short term to solve the pb of uploading data into datastore during provisioning.

jfomhover (approved):

PR #73, opened 2022-09-23
Simplify open sandbox setup script

jfomhover:
Since this is purely for demo purpose, some of the choices in this setup script can be simplified.

Also did some clean-up and documentation of the parameters.

Before merging:
- [x] run the sample experiment out of the box


garg-amit (approved):
Looks good to me!

PR #86, opened 2022-09-27
Make each provisioning script pick-and-choose (can run independently)

jfomhover:
This makes the bicep provisioning scripts more independent from one another, so that users can run one and not the others.

In particular, it splits orchestrator into 1) creating the azureml workspace and 2) create compute+storage for orchestrator.

The scripts also have the capacity to use either SystemAssigned or UserAssigned identities in a parameter switch.

The UAI version is experiencing issues for the moment, but the MSI version works quite well.


thomasp-ms:
@jfomhover just curious, why this location logic [here](https://github.com/Azure-Samples/azure-ml-federated-learning/blob/e771c3838dc75e0edd25b76c6e5864d65cba42c2/mlops/bicep/modules/resources/open_azureml_workspace.bicep#L93)?

thomasp-ms:
@jfomhover I see [here](https://github.com/Azure-Samples/azure-ml-federated-learning/blob/e771c3838dc75e0edd25b76c6e5864d65cba42c2/mlops/bicep/modules/resources/open_azureml_workspace.bicep#L57) for the orchestrator storage account you're using the `RAGRS` SKU. Any particular reason for using this _geo-replicated_ variant?

I know this account will (in theory) only store models so the location of the data shouldn't matter, but I figured the `LRS` variant might be a bit safer...

thomasp-ms:
@jfomhover [This part](https://github.com/Azure-Samples/azure-ml-federated-learning/blob/e771c3838dc75e0edd25b76c6e5864d65cba42c2/mlops/bicep/modules/orchestrators/open_orchestrator_blob.bicep#L86) (stating the container depends on the storage account) triggers a warning. I do not understand why, this seems counter-intuitive to me. I wonder if we should dig further into this, maybe it is a signal that there is something amiss somewhere?

jfomhover:
> @jfomhover just curious, why this location logic [here](https://github.com/Azure-Samples/azure-ml-federated-learning/blob/e771c3838dc75e0edd25b76c6e5864d65cba42c2/mlops/bicep/modules/resources/open_azureml_workspace.bicep#L93)?



> @jfomhover just curious, why this location logic [here](https://github.com/Azure-Samples/azure-ml-federated-learning/blob/e771c3838dc75e0edd25b76c6e5864d65cba42c2/mlops/bicep/modules/resources/open_azureml_workspace.bicep#L93)?

I took this logic from a quickstart example, I am not sure I think it is probably because app insights doesn't cover all regions?

jfomhover:
> @jfomhover I see [here](https://github.com/Azure-Samples/azure-ml-federated-learning/blob/e771c3838dc75e0edd25b76c6e5864d65cba42c2/mlops/bicep/modules/resources/open_azureml_workspace.bicep#L57) for the orchestrator storage account you're using the `RAGRS` SKU. Any particular reason for using this _geo-replicated_ variant?
> 
> I know this account will (in theory) only store models so the location of the data shouldn't matter, but I figured the `LRS` variant might be a bit safer...

yes please downgrade to LRS, thanks for spotting this

thomasp-ms:
> 

fair enough

PR #104, opened 2022-10-06
Integrate subgraphs - native pipeline

### LINKED ISSUE #102: Integrate Subgraph
- It's challenging to navigate various FL components in UI, especially when the number of iterations exceeds 10.

**Describe the solution you'd like**
- Integration of subgraph components would allow switching b/w different components easily as well as would provide a clear view of the pipeline in UI.

**Describe alternatives you've considered**
- Earlier DPv1.5 components were used to deal with this problem. As we are migrating everything to DPv2, it's better to have an entire solution in DPv2.

garg-amit:
## Purpose

* Integrate subgraphs with the native code. 

closes #102 

```
[x] Yes
[ ] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

Submit a native pipeline and go to the azure studio to look at subgraphs.


thomasp-ms:
@garg-amit this is not a full review yet, but I noticed something odd in the submitted job. You seem to be including the pre-processing step in every subgraph. We don't want that, even if the preprocessing results are being reused and the steps aren't really rerun; that would just be confusing to the user.

I suggest keeping preprocessing steps separate, and have the subgraph only contain silo training and aggregation.

garg-amit:
Thanks for the initial review, @thomasp-ms. Unfortunately, the subgraph component does not support **kwargs. That being said, if we keep the preprocessing component outside the subgraph then we have to statically define the subgraph function with the exact no. of inputs parameters i.e. 2*num_silos + other params so a user has to change the code each time when he/she wants to add/remove a silo (it won't be a just config change anymore). Let me know if you think otherwise or have a workaround in mind.

examples/pipelines/fl_cross_silo_native/submit.py
```diff
@@ -23,6 +23,8 @@
 # to handle yaml config easily
 from omegaconf import OmegaConf
 
+# subgraph
+os.environ["AZURE_ML_CLI_PRIVATE_FEATURES_ENABLED"] = "true"
```
thomasp-ms, line 27: If we have to do this, does this mean it's OK to put in a public repo?

examples/pipelines/fl_cross_silo_native/submit.py
```diff
@@ -204,93 +205,89 @@ def fl_cross_silo_internal_basic():
...
+        silo_training_step.name = f"silo_{silo_index}_training"
+
+        # make sure the compute corresponds to the silo
+        silo_training_step.compute = silo_config.compute
 
         # make sure the data is written in the right datastore
-        aggregate_weights_step.outputs.aggregated_output = Output(
-            type=AssetTypes.URI_FOLDER,
+        silo_training_step.outputs.model = Output(
+            type=AssetTypes.CUSTOM_MODEL,
```
thomasp-ms, line 239: Are we sure we want to use CUSTOM_MODEL here? I thought that was triggering the registration of the model somewhere else, which we don't want. (sorry if I'm late to the party and this has been dealt with already)

thomasp-ms:
> Thanks for the initial review, @thomasp-ms. Unfortunately, the subgraph component does not support **kwargs. That being said, if we keep the preprocessing component outside the subgraph then we have to statically define the subgraph function with the exact no. of inputs parameters i.e. 2*num_silos + other params so a user has to change the code each time when he/she wants to add/remove a silo (it won't be a just config change anymore). Let me know if you think otherwise or have a workaround in mind.

Hmmm, this is unfortunate. If what you're saying about **kwargs not being supported is true, then I agree your approach with repeated preprocessing steps is better than forcing the user to re-write code for different numbers of silos.

So far I have no workaround in mind besides the obvious "define 2*MaxNumberOfSilos optional inputs", since we do have an arbitrary limit on the max number of silos we want to support (100). However this workaround is bad for many reasons so let's stick with your solution.

Let's chat more about this during our next sync.

thomasp-ms (approved):
One minor comment about exposing private features, and one less minor about a data type. The rest looks good.

Since **kwargs are not supported, I agree with your approach of repeating the pre-processing steps. Might be slightly confusing to the user, but better than having to write code every time they want to change the number of silos.

examples/pipelines/fl_cross_silo_native/submit.py
garg-amit: my bad! That should be a `uri_folder`

examples/pipelines/fl_cross_silo_native/submit.py
garg-amit: That's a good question. Let me cc @jfomhover (he might have an opinion on the same)

garg-amit:
Integrating subgraphs in just the factory code for now since subgraphs are in the experimental phase.

PR #107, opened 2022-10-09
Fix and clean-up of open sandbox provisioning scripts, align submit scripts

jfomhover:
## Purpose

This PR fixes the open sandbox setup, and cleans up the bicep scripts for the open non-secure provisioning.

In particular:
- orchestrator and silos are now just a "pair" of compute and storage with the same provisioning script
- permission model is applied outside of pair provisioning
- fixes the issue with cross-geo computes

```
[x] Yes
[ ] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## What to Check

* provision an open sandbox with bicep scripts
* run both native+factory submits out of the box


mlops/bicep/open_sandbox_setup.bicep
```diff
@@ -43,12 +46,6 @@ param siloRegions array = [
 @description('The VM used for creating compute clusters in orchestrator and silos.')
 param computeSKU string = 'Standard_DS13_v2'
 
-@description('The number of nodes used for creating compute clusters in orchestrator.')
-param computeNodes int = 4
```
thomasp-ms, line 47: I suppose this deletion was due to a wrong merge. This is fairly minor so let's stick with the magic number - I'll fix it in the next iteration

thomasp-ms (approved):

PR #113, opened 2022-10-11
Update quickstart to mention rg clean-up

jfomhover:
Title says it all.

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -56,7 +56,6 @@ We will provision:
 
     > Notes:
       > - If someone already provisioned a demo with the same name in your subscription, change `demoBaseName` parameter to a unique value.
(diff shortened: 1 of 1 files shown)


docs/quickstart.md
```diff
@@ -37,18 +37,24 @@ We will provision:
     az account set --name <subscription name>
     ```
 
-2. Run the bicep deployment script:
+2. Optional: Create a new resource group for the demo resources. Having a new group would make it easier to delete the resources afterwards.
```
garg-amit, line 40: Should we add the following statement explicitly - Deleting the resource group would delete all resources created by this script?

thomasp-ms (changes requested):

garg-amit (approved):

PR #127, opened 2022-10-21
Set allowBlobPublicAccess to False in all storageAccount provisioning

jfomhover:
This is intended to avoid having someone create a public container in those.

TODO:
- [ ] fix the diff test
- [ ] rebuild biceps


jfomhover:
Abandoned, this will be covered as part of the refactoring done in #132

PR #138, opened 2022-10-25
Add "open aks with cc" provision tutorial and bicep scripts

jfomhover:
## Purpose

This adds an experimental provisioning tutorial + bicep to create an AKS cluster with confidential compute nodes, Not yet secured with vnet and private endpoints (future work).

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

### DIFF
--- a/mlops/bicep/modules/azureml/attach_aks_training_to_azureml.bicep
+++ b/mlops/bicep/modules/azureml/attach_aks_training_to_azureml.bicep
@@ -1,0 +1,61 @@
+// This BICEP script will attach an AKS cluster
+// to a given AzureML workspace for training (NOT inferencing).
+
+// resource group must be specified as scope in az cli or module call
+targetScope = 'resourceGroup'
+
+// required parameters
+@description('Name of AzureML workspace to attach compute+storage to.')
+param machineLearningName string
+
+@description('The region of the machine learning workspace')
+param machineLearningRegion string = resourceGroup().location
+
+@description('Resource ID of the AKS cluster.')
+param aksResourceId string
+
+@description('Region of the AKS cluster.')
+param aksRegion string
+
+@description('How to name this compute in Azure ML')
+param amlComputeName string
+
+@description('Name of the existing UAI for the compute cluster.')
+param computeUaiName string
+
+// provision a user assigned identify for this silo
+resource uai 'Microsoft.ManagedIdentity/userAssignedIdentities@2022-01-31-preview' existing = {
+  name: computeUaiName
+  scope: resourceGroup()
+}
+
+var identityPrincipalId = uai.properties.principalId
+var userAssignedIdentities = {'/subscriptions/${subscription().subscriptionId}/resourceGroups/${resourceGroup().name}/providers/Microsoft.ManagedIdentity/userAssignedIdentities/${uai.name}': {}}
+
+resource workspace 'Microsoft.MachineLearningServices/workspaces@2022-05-01' existing = {
+  name: machineLearningName
+  scope: resourceGroup()
+}
+
+// attach the AKS cluster to the workspace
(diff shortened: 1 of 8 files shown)


mlops/bicep/modules/computes/open_new_aks_with_confcomp.bicep
```diff
@@ -0,0 +1,149 @@
...
+param aksClusterName string
+
+@description('How to name this compute in Azure ML')
+param amlComputeName string = aksClusterName
+
+@description('Specifies the location of the compute resources.')
+param computeRegion string
+
+@description('Optional DNS prefix to use with hosted Kubernetes API server FQDN.')
+@maxLength(54)
```
thomasp-ms, line 29: great find, I didn't know about this maxlength functionality, or about the min/maxValue

mlops/bicep/modules/azureml/attach_aks_training_to_azureml.bicep
```diff
@@ -0,0 +1,61 @@
...
+param computeUaiName string
+
+// provision a user assigned identify for this silo
+resource uai 'Microsoft.ManagedIdentity/userAssignedIdentities@2022-01-31-preview' existing = {
+  name: computeUaiName
+  scope: resourceGroup()
+}
+
+var identityPrincipalId = uai.properties.principalId
+var userAssignedIdentities = {'/subscriptions/${subscription().subscriptionId}/resourceGroups/${resourceGroup().name}/providers/Microsoft.ManagedIdentity/userAssignedIdentities/${uai.name}': {}}
```
thomasp-ms, line 33: just a nit, but I think we're doing this very same string interpolation at least 4 times, if I'm counting correctly. Do you know if there's a way we can introduce a custom function in bicep, that we could call in different bicep files? (I did a quick search and didn't find anything, but maybe you know a way already? If not, just drop it - this is neither critical nor expected to change)

mlops/bicep/modules/azureml/deploy_aks_azureml_extension_via_script.bicep
```diff
@@ -0,0 +1,122 @@
...
+
+// @description('Extension authors can publish versions in different release trains such as Stable, Preview, etc.')
+// param releaseTrain string = 'Stable'
+
+// @description('indicates the namespace within which the release is to be created. This parameter is only relevant if scope parameter is set to cluster.')
+// param releaseNamespace string = 'azureml'
+
+resource aks 'Microsoft.ContainerService/managedClusters@2022-05-02-preview' existing = {
+  name: clusterName
+  scope: resourceGroup()
```
thomasp-ms, line 66: I don't see this scope line in the other deploy file, but maybe I'm missing something?

thomasp-ms (approved):

mlops/bicep/modules/azureml/attach_aks_training_to_azureml.bicep
jfomhover: I think I might actually have found a way using resource() instead, which implicitely builds that whole path given a type and resource ID. I'll file an issue to implement in a next round.

jfomhover: https://github.com/Azure-Samples/azure-ml-federated-learning/issues/143

mlops/bicep/modules/azureml/deploy_aks_azureml_extension_via_script.bicep
jfomhover: scope is implicit when not specified, and is set to the scope of the deployment itself (specified as targetScope on top of the file). so technically this line here is not necessary at all but it enforces the idea that the aks should be located in the same resource group we're deploying here.

PR #140, opened 2022-10-25
Add Bicep build vs ARM template diff test 

garg-amit:
## Purpose

* Diff Test - Bicep build vs ARM template 
* Fix bugs related to incorrect triggers of provisioning tests

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```


.github/actions/bicep-to-arm-template-diff/action.yaml
```diff
@@ -0,0 +1,36 @@
...
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: Build bicep into arm
+      shell: bash
+      run: |
+        az bicep build --file ${{ inputs.source-file-path }} --stdout | jq -S . > source.json
+        cat source.json
```
jfomhover, line 18: this is going to generate a lot of logs, I think it's fine to not cat the entire ARM since we have the diff below, and if something is wrong we could just reproduce locally in the branch?

jfomhover (approved):
This is great to have! I'm worried about having cat in the action though.

.github/actions/bicep-to-arm-template-diff/action.yaml
garg-amit: oh my bad, I did that when I was debugging. Let me quickly fix that

PR #150, opened 2022-10-31
Implement option to turn orchestrator storage fully private (behind PLE)

jfomhover:
## Purpose

This PR implements multiple changes to the code to allow for orchestrator storage to be fully private. In order to support this, changes are:
* add option in the publicip sandbox bicep
* reuse single private endpoint code accross all provisioning scripts
* enforce static IP address for orchestrator and silo endpoints
* make all private DNS zone names customizable
* set A record with all IP adresses in private DNS zone

NOTE: when creating multiple endpoints in multiple vnets using the same private DNS zone, the IP address of each endpoint will overwrite the previous one. In this PR we are using static IP adresses so that we can create a unique record in the DNS zone with all the IP adresses from each vnet (orch + silos).

Not from the user experience perspective.

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[x] Code style update (formatting, local variables)
[ ] Other... Please describe:
```

### DIFF
--- a/mlops/bicep/modules/azureml/open_azureml_workspace.bicep
+++ b/mlops/bicep/modules/azureml/open_azureml_workspace.bicep
@@ -8,8 +8,11 @@
 targetScope = 'resourceGroup'
 
 // required parameters
+@description('Base name to create all the resources')
+param baseName string
+
 @description('Machine learning workspace name')
-param machineLearningName string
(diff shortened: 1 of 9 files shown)


thomasp-ms (approved):
Looks good overall, but I'll be honnet: some of it went way over my head.

PR #164, opened 2022-11-07
Pneumonia xray example

### LINKED ISSUE #151: Provide example FL job for medical imaging application
We want to propose example jobs that are relevant to various industries.

**Describe the solution you'd like**
- Use the chest xray dataset from kaggle
- Port Harmke's code (which is using this dataset, but also using NVFlare - we'll want to get rid of that last part)

thomasp-ms:
**This PR adds an example cross-silo FL pipeline for doing pneumonia detection on native Azure ML.** It is directly adapted from [that example using NVFlare](https://github.com/Azure/medical-imaging/tree/main/federated-learning). We use Azure ML computes clusters and Azure storages in various regions to mimic the case of geographically distributed data.

The entry point for running this example is the [pneumonia.md](https://github.com/Azure-Samples/azure-ml-federated-learning/blob/thomas/pneumonia-xray-example/docs/real-world-examples/pneumonia.md) file. It contains all the instructions. Here is a quick summary to help with the code review.

1. Provisioning - For simplicity, we recommend creating an open setup, just like in the quickstart.
2. Data - Get the dataset from Kaggle and split it into 3 silos by cloning the repo, adding some GitHub secrets and running a GitHub action.
3. Job - Modify the config file to point at your workspace, then run a single command.

Successful run [here](https://ml.azure.com/runs/tender_cord_v9lkcc23ry?wsid=/subscriptions/48bbc269-ce89-4f6f-9a12-c6f91fcb772d/resourcegroups/thopo-owner-fldev-rg/workspaces/aml-tp-pneumo&tid=72f988bf-86f1-41af-91ab-2d7cd011db47#).

Closes #151 

## Purpose

* The goal with this addition is to increase the advertisement value of our repo. We are targeting the healthcare industry.

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
* Get the code by **forking** the repo (not _cloning_, see more details in the instructions).
* Test the code by following the instructions in the [pneumonia.md](https://github.com/Azure-Samples/azure-ml-federated-learning/blob/thomas/pneumonia-xray-example/docs/real-world-examples/pneumonia.md) doc.

> Note: It is not expected of you to do the whole testing described above as part of the code review. This is more for the bug bash. For the code review, it is totally OK if you just read the code and not actually run it. But if you do want to run it, don't let this note stop you ;)

## What to Check
Verify that the following are valid:
* the data are split properly into the 3 silos;
* the FL job runs successfully.

## Follow-up work
- [ ] use a shared component for model aggregation
- [ ] only keep one environment for all examples
- [ ] add this experiment to the CI/CD?
- [ ] revisit the story around data provisioning


examples/pipelines/pneumonia/environment.yml
```diff
@@ -0,0 +1,8 @@
+name: fl_pneumonia_env
```
jfomhover, line 1: can we name this file conda.yml?

docs/real-world-examples/pneumonia.md
```diff
@@ -0,0 +1,54 @@
...
+
+1. Create a service principal with contributor access to your resource group by running the following Azure CLI command.
+```bash
+  az ad sp create-for-rbac --name "<service-principal-name>" --role contributor --scopes /subscriptions/<subscription-id>/resourceGroups/<your-resource-group-name> --sdk-auth
+```
+2. Create a GitHub secret in your _forked_ repository. Name it AZURE_CREDENTIALS_DATAPREP and copy and paste the output of the above command to the Value field of the secret.
+3. Add the following secrets to your repository with corresponding values:
+  - KAGGLE_USERNAME: your Kaggle username.
+  - KAGGLE_KEY: your Kaggle API key. More info [here](https://www.kaggle.com/docs/api).
+4. Navigate to the GitHub Actions tab, and run the workflow with the name *FL data preparation - pneumonia example*. Provide the names of your workspace and resource group. This will register the datasets required for your experiment in your workspace.
```
jfomhover, line 41: I think this line is from Harmke's previous tutorial, but doesn't apply here.

docs/real-world-examples/pneumonia.md
```diff
@@ -0,0 +1,54 @@
...
+
+Each phase is described in the subsections below.
+
+### Provisioning
+This phase is easy. Just follow the instructions in the [quickstart](../quickstart.md) to provision an open sandbox. Make note of the name of the resource group you provisioned, as well as the name of the workspace.
+
+### Data preparation
+The goal of this phase is to upload to Azure ML 3 distinct datasets (one per silo), each subdivided into train, test, and validation. This is achieved by running a GitHub action that will download the dataset from [Kaggle](https://www.kaggle.com/datasets/paultimothymooney/chest-xray-pneumonia), split it evenly between the silos, and upload the 3 data assets. We will need to first create an Azure Service Principal with enough permissions to create the data assets in Azure ML. The whole procedure is explained below, and closely follows [the one](https://github.com/Azure/medical-imaging/blob/main/federated-learning/README.md#1-prepare-your-experiment) developed by Harmke Alkemade _et al_ .
+
+1. Create a service principal with contributor access to your resource group by running the following Azure CLI command.
```
jfomhover, line 33: I would rather provide a job in yaml that users can run in their own workspace, it would completely save the step to create a fork, then an SP, etc.

Basically, you could provide something like this:
https://github.com/Azure/azureml-assets/blob/main/benchmark/src/jobs/create_places2_dataset.yml

and then just ask users to run `az ml job create` with that job first, it would unpack the dataset on each silo

docs/real-world-examples/pneumonia.md
```diff
@@ -0,0 +1,54 @@
...
+```bash
+  az ad sp create-for-rbac --name "<service-principal-name>" --role contributor --scopes /subscriptions/<subscription-id>/resourceGroups/<your-resource-group-name> --sdk-auth
+```
+2. Create a GitHub secret in your _forked_ repository. Name it AZURE_CREDENTIALS_DATAPREP and copy and paste the output of the above command to the Value field of the secret.
```
jfomhover, line 37: This whole process applies to Harmke's repository but I don't think this process should apply in our case. I would rather give our users some job to run than have to mess up with github secrets etc.

jfomhover: Feel free to push only the pipeline part of your job, that assumes the data is in a given location in each silo, then you can push the job to upload the data in a distinct PR

examples/pipelines/pneumonia/submit.py
```diff
@@ -0,0 +1,302 @@
...
+####################################
+
+# Loading the component from their yaml specifications
+training_component = load_component(
+    source=os.path.join(COMPONENTS_FOLDER, "traininsilo", "spec.yaml")
+)
+
+aggregate_component = load_component(
+    source=os.path.join(
+        COMPONENTS_FOLDER, "aggregatemodelweights", "spec.yaml"
```
jfomhover, line 137: What I meant this morning is that it would be easy to divert just this line to call the generic aggregatemodelweights instead. But up to you, as long as we modify this before the release.

jfomhover (changes requested):

docs/real-world-examples/pneumonia.md
thomasp-ms: I doubt this is going to be much easier since we're also going to have to deal with some sort of secrets for the Kaggle credentials, but sure I'll try to use an Azure ML command job instead

docs/real-world-examples/pneumonia.md
thomasp-ms: People will have to deal with secrets no matter what, as long as we're using this kaggle dataset. 

I'd rather put everything in a single PR

thomasp-ms: Hmmm, actually with the uncertainty about how to do the data provisioning, I'd rather get this in soon, then do a follow-up the data provisioning

garg-amit (approved):

docs/real-world-examples/pneumonia.md
ShizeSu06, line 37: Minor suggestion: maybe it helps to add a hyperlink here (e.g., https://github.com/Azure/actions-workflow-samples/blob/master/assets/create-secrets-for-GitHub-workflows.md)? Some 3P customers might be less/not familiar with such github secrets and actions, and a hyperlink might help.

examples/components/PNEUMONIA/traininsilo/conda.yaml
```diff
@@ -0,0 +1,14 @@
+name: pneumonia_train_conda_env
+channels:
+  - defaults
+  - pytorch
+dependencies:
+  - python=3.7.11
+  - pip=20.0
```
ShizeSu06, line 7: Nit: pip=20.0 seems to be quite old/outdated. Is there any particular reason to use this old version pip? Or you just got this from Harmke's example and thus didn't spend time adjusting it? 

Might be slightly better to use less outdated version, but it is a tiny issue, and I'm ok if you want to merge the PR ASAP as it is.

ShizeSu06 (approved):

docs/real-world-examples/pneumonia.md
thomasp-ms: Good suggestion, will do

examples/components/PNEUMONIA/traininsilo/conda.yaml
thomasp-ms: You're making a good point here too. I have no reason other than "this is what we were using for MNIST traininsilo" :) 

I have created [an issue](https://github.com/Azure-Samples/azure-ml-federated-learning/issues/180) for fixing this in a subsequent PR.

jfomhover (approved):

PR #177, opened 2022-11-17
Named Entity Recognition example

garg-amit:
## Purpose

* This PR showcases how to train a federated model for the Named Entity Recognition task. This tutorial uses the [MutliNERD](https://github.com/Babelscape/multinerd/blob/master/README.md) dataset.
* This [doc](https://github.com/Azure-Samples/azure-ml-federated-learning/blob/gargamit/nlp-use-case/docs/real-world-examples/ner.md) describes a step-by-step process for carrying out this experiment, including provisioning resources, uploading data, creating training scripts, etc. Any NLP work can be carried out in a federated fashion using this example as a template.
* To bifurcate and upload data to multiple silos, here is a standalone [CommandJob](https://github.com/Azure-Samples/azure-ml-federated-learning/tree/gargamit/nlp-use-case/examples/components/NER/upload_data_to_silos) script.

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code (branch-name: nlp-use-case)

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

Follow the instructions provided in the [doc](https://github.com/Azure-Samples/azure-ml-federated-learning/blob/gargamit/nlp-use-case/docs/real-world-examples/ner.md).

## What to Check
* The data is successfully uploaded to silos.
* Pipeline is completed successfully.


examples/components/NER/aggregatemodelweights/spec.yaml
```diff
@@ -0,0 +1,39 @@
+$schema: https://azuremlschemas.azureedge.net/latest/commandComponent.schema.json
```
majercakdavid, line 1: Is it possible to just reuse weights aggregation script from the MNIST directory?

examples/pipelines/fl_cross_silo_literal/submit.py
```diff
@@ -51,9 +51,9 @@
 parser.add_argument(
     "--example",
     required=False,
-    choices=["MNIST", "HELLOWORLD"],
+    choices=["MNIST", "HELLOWORLD", "NER"],
```
majercakdavid, line 54: I believe we do not want to alter files not related to example

majercakdavid (changes requested):

examples/components/NER/aggregatemodelweights/spec.yaml
thomasp-ms: I will fix that part in my [upcoming PR](https://github.com/Azure-Samples/azure-ml-federated-learning/pull/182), after this one has gone in

examples/pipelines/fl_cross_silo_literal/submit.py
thomasp-ms: I'm not shocked by this. If the generic examples also work, we might as well have them do the NER task. **IF** that's happening though, why only with literal and not native?

garg-amit: my bad! The literal/factory `submit.py` is pretty much the same for the `NER` example but the `config.yml` is different. So I need to remove this from here.

examples/components/NER/traininsilo/run.py
```diff
@@ -0,0 +1,483 @@
...
+        self._batch_size = batch_size
+        self._experiment_name = experiment_name
+        self._iteration_num = iteration_num
+        self._model_path = model_path
+
+        # device
+        self.device_ = torch.device("cuda:0" if torch.cuda.is_available() else "cpu")
+
+        # tokenizer
+        model_name = "bert-base-cased"
```
thomasp-ms, line 69: This seems to be hardcoded in the pre-processing component too. For me it's okay to merge as it is, but we might want to take a note of that potential slight improvement in the future (define the model type as an input parameter, or something hardcoded in a single location in the submit.py file).

thomasp-ms (commented):
Looks correct overall. I didn't dive into all details of the training; the metrics in the successful job convinced me that the training indeed is working.

docs/real-world-examples/ner.md
```diff
@@ -0,0 +1,53 @@
...
+
+### Provision Resources
+The instructions are provided in the [quickstart](../quickstart.md) to provision an open sandbox. Make note of the name of the resource group you provisioned, as well as the name of the workspace.
+
+### Upload Data
+The steps to upload data to various datastores are as follows:
+
+Note: This is not required if you've already uploaded the data into their respective datastores.
+
+1. Make sure the train and test paths in the `./examples/components/NER/upload_data_to_silos/spec.yaml` and `./examples/components/NER/preprocessing/spec.yaml` files are the same.
```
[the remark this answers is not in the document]
garg-amit: oops! These files `./examples/components/NER/upload_data_to_silos/spec.yaml` and `./examples/pipelines/ner/config.yaml` should be consistent. Thanks for pointing that out, Thomas!

examples/components/NER/traininsilo/run.py
garg-amit: Did the required changes to make tokenizer-name and model-name configurable.

examples/components/NER/upload_data_to_silos/script.py
```diff
@@ -0,0 +1,101 @@
+import argparse
```
majercakdavid, line 1: maybe rename file to run.py to align with rest of the folders?

examples/components/NER/upload_data_to_silos/spec.yaml
```diff
@@ -0,0 +1,40 @@
...
+    path: azureml://datastores/datastore_silo1_francecentral/paths/multinerd/train/
+  silo1_test_data:
+    type: uri_folder
+    mode: mount
+    path: azureml://datastores/datastore_silo1_francecentral/paths/multinerd/test/
+
+  silo2_train_data:
+    type: uri_folder
+    mode: mount
+    path: azureml://datastores/datastore_silo2_brazilsouth/paths/multinerd/train/
```
majercakdavid, line 25: Does this actually work? As per my understanding according to discussion with @jfomhover if an orchestrator is able to write to silo storage this is a bug.

garg-amit: This works for the open setup but we still need to figure out how to do this for a secure setup. IMO, this situation would never come in a real scenario/production where we need to bifurcate data and then upload it to multiple silos. If we already have data in one place, then that's no more an FL problem. The other scenario would be if the data is already partitioned then we can provide a script later that would run in silo computes and upload data to their respective datastores.

majercakdavid: This is interesting - how is it possible that it works with following UID on orchestrator?
image
It should not have access to other blobstorages. Do you know why is that happening?

garg-amit: This might be happening because I manually created GPU computes and all the data stores are under one storage account. Happy to elaborate on this on Teams if needed!

majercakdavid: The problem with such setup though is that storage account has assigned location and thus all the data will be hosted in the same region, right? Unless I may be missing something?

Teams is also a way but as this is something we all need to solve for industrial examples, it might be better to have it in a shared space.

majercakdavid: Follow up: Discussed in a call with a conclusion that we will try to create a compute that can access all the blob storages and the data will be uploaded to corresponding regions.

docs/real-world-examples/ner.md
```diff
@@ -0,0 +1,53 @@
...
+  - Note that to set permissions, you typically need _Owner_ role in the subscription or resource group - _Contributor_ role is not enough. This is key for being able to _secure_ the setup.
+- [install the Azure CLI](https://learn.microsoft.com/en-us/cli/azure/install-azure-cli).
+
+## Procedure
+Please follow the below instructions to provision resources, upload data, and then run the experiment:
+
+### Provision Resources
+The instructions are provided in the [quickstart](../quickstart.md) to provision an open sandbox. Make note of the name of the resource group you provisioned, as well as the name of the workspace.
+
+### Upload Data
```
ShizeSu06, line 21: One thing that is the same across all three examples is that, the data-upload section is only applicable for "running the examples" and not expected applicable for real FL scenario, and they are provided here simply for the only purpose of convenience of running the FL example for the customer, and users should ignore this section when they want to do their real FL experiment for their scenario. (In real FL scenario, the data from all silos would never be in a single central location, and the 'data-uploading' section in all the three examples are not expected to be applicable here; instead, each silo owner are expected to upload there own private data to their own private & secure datastores.) It might be helpful to explicitly point this out somewhere, maybe under docs/README.md, to avoid any potential confusions for the 3P customers.

ShizeSu06 (approved):

docs/README.md
```diff
@@ -61,6 +61,9 @@ In addition to the [literal](../examples/pipelines/fl_cross_silo_literal/) and [
 ### Pneumonia detection from chest radiographs
 In this example, we train a model to detect pneumonia from chest radiographs. The model is trained on a dataset from the [NIH Chest X-ray dataset](https://www.kaggle.com/nih-chest-xrays/data). This example is adapted from [that solution](https://github.com/Azure/medical-imaging/tree/main/federated-learning) by Harmke Alkemade _et al._ See [here](./real-world-examples/pneumonia.md) for detailed instructions on how to run this example.
 
+### Named Entity Recognition using MultiNERD dataset
```
ShizeSu06, line 64: @jfomhover  Currently all the 3 real world fl examples all only provide literal submission script (i.e., not using pipeline subgraph, not using fl-factory). Do we want to provide submission script that leverage pipeline subgraph and/or fl factory here? Or you prefer to only keep the literal submission script for these 3 real world examples for now?

docs/real-world-examples/ner.md
garg-amit: Added a note in the Readme

examples/components/NER/upload_data_to_silos/spec.yaml
garg-amit: Created a compute that has access to the silos' storage accounts and updated the documentation accordingly.

thomasp-ms (approved):

examples/pipelines/ner/environment.yml
```diff
@@ -0,0 +1,10 @@
+name: fl_ner_env
```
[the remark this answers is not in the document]
garg-amit: oh yeah let me remove this one

docs/README.md
jfomhover: let's stick to literal for now

PR #188, opened 2022-11-28
Provisioning script to create cpu+gpu computes  

### LINKED ISSUE #169: Extend provisioning and factory to support cpu+gpu computes

In our pipelines, we want to support having both cpu and gpu computes (preprocessing in cpu, training in gpu).
The request is to add gpu provisioning as part of the silo compute/datastore pairs. It would also require to be able to specify multiple compute targets in the literal/factory submit codes.

garg-amit:
## Purpose

* Provide a script to create cpu+gpu computes sharing the same datastore.
* Modify both open and vnet setup to incorporate these changes.
* Generalize the names of the resources and do required changes in the pipeline's configs.
* Fix token expiration issue when provisioning resources.
* List all jobs when deleting history runs.

closes #169 

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

## What to Check
* Check the CI/CD tests.


mlops/bicep/vnet_publicip_sandbox_setup.bicep
```diff
@@ -161,12 +170,15 @@ module silos './modules/fl_pairs/vnet_compute_storage_pair.bicep' = [for i in ra
     pairRegion: siloRegions[i]
     tags: tags
 
-    pairBaseName: '${demoBaseName}-silo${i}-${siloRegions[i]}'
+    pairBaseName: '${demoBaseName}-silo${i}'
 
-    computeName: 'cpu-silo${i}-${siloRegions[i]}' // let's not use demo base name
-    computeSKU: computeSKU
+    compute1Name: 'silo${i}-01' // let's not use demo base name in cluster name
```
jfomhover, line 175: could we keep the region here? it looks much nicer

proposal: `silo${i}-${siloRegions[i]}-1`

jfomhover (approved):
Overall looks good, but I'd love to keep the region name in the silo compute+datastore, it provides a more fancy naming convention for demos

mlops/bicep/vnet_publicip_sandbox_setup.bicep
garg-amit: I removed the region name to avoid one additional step. Users can have different region-quota limits. If a user changes regions while provisioning resources, the respective names need to be replaced(YAML files) if the region is a part of the compute/datastore name. Please let me know if you feel otherwise.

jfomhover: Yes you are right if someone needs to adapt the provisioning, they might have to tweak their config files. But what I loved about the previous names is that out-of-the-box effect of our demo, showing regional names in computes, and in the metrics. We could figure out a way to address that in another setting somewhere else though.

garg-amit: As discussed, I'll create another PR to accommodate these changes.

PR #194, opened 2022-12-02
Micro-fixed after bug bashing all 3 industry examples

jfomhover:
## Purpose

* pre-validate our 3 real world scenarios to ensure usability

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

### DIFF
--- a/examples/components/CCFRAUD/preprocessing/conda.yaml
+++ b/examples/components/CCFRAUD/preprocessing/conda.yaml
@@ -13,4 +13,3 @@ dependencies:
     - azureml-core
     - pandas
     - scikit-learn
-    - hydra-core
--- a/examples/components/CCFRAUD/preprocessing/spec.yaml
+++ b/examples/components/CCFRAUD/preprocessing/spec.yaml
@@ -36,4 +36,4 @@ command: >-
 
 environment: 
   conda_file: ./conda.yaml
-  image: mcr.microsoft.com/azureml/openmpi3.1.2-ubuntu18.04
+  image: mcr.microsoft.com/azureml/openmpi4.1.0-ubuntu20.04
--- a/examples/components/CCFRAUD/preprocessing/run.py
+++ b/examples/components/CCFRAUD/preprocessing/run.py
@@ -7,7 +7,6 @@ import numpy as np
 from sklearn.preprocessing import StandardScaler
 import pandas as pd
 import mlflow
-import hydra
 
 from azureml.core import Run, Workspace
 
--- a/examples/components/CCFRAUD/traininsilo/spec.yaml
+++ b/examples/components/CCFRAUD/traininsilo/spec.yaml
@@ -74,4 +74,4 @@ command: >-
   $[[--batch_size ${{inputs.batch_size}}]]
 environment: 
   conda_file: ./conda.yaml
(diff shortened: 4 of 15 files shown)


examples/components/CCFRAUD/preprocessing/conda.yaml
```diff
@@ -1,16 +0,0 @@
...
-  - defaults
-  - pytorch
-dependencies:
-  - python=3.8
-  - pip=22.2.2
-  - pytorch=1.12.1
-  - torchvision=0.13.1
-  - cudatoolkit=11.3
-  - pip:
-    - azureml-mlflow
```
majercakdavid, line 12: I assume this file should not have been deleted, right?

examples/components/CCFRAUD/preprocessing/spec.yaml
```diff
@@ -1,10 +1,10 @@
...
 $schema: https://azuremlschemas.azureedge.net/latest/commandComponent.schema.json
-name: preprocessing_in_silo
-version: 0.1.12
-display_name: Pre-Processing (in silo)
+name: ccfraud_preprocessing_in_silo
+version: 0.3.0
+display_name: CC Fraud Pre-Processing (in silo)
 type: command
-description: Component for preprocessing raw data in a given silo. The dataset is downloaded and stored in the blob storage
+description: Component for preprocessing raw data from the ccfraud dataset.
```
majercakdavid, line 7: Labeling this as "ccfraud dataset" could be misleading as this is not really our dataset. I would maybe suggest something along the lines of "Component for preprocessing raw data from silo's blob storage"

examples/pipelines/ccfraud/submit.py
```diff
@@ -181,7 +187,7 @@ def getUniqueIdentifier(length=8):
 @pipeline(
     description=f'FL cross-silo basic pipeline and the unique identifier is "{pipeline_identifier}" that can help you to track files in the storage account.',
 )
-def fl_cross_silo_internal_basic():
+def fl_ccfraud_basic():
```
majercakdavid, line 190: Why to rename function in every example to be custom one? Isn't it better to name them same way so we have some consistency?

examples/components/CCFRAUD/preprocessing/conda.yaml
jfomhover: actually I deleted it intentionally, in the training scripts, there are very minimal dependencies that are all present in one of the AzureML curated environment. Using a curated environment shortcuts the build step and skips 10 mins of run time

examples/pipelines/ccfraud/submit.py
jfomhover: The name of the function is used (currently) as a run name. So my intention here is we have 3 different experiments, one for each scenario, and in each experiment, we have runs with name aligning with the scenario as well so that when you look at "all runs" in the workspace, you see which one is which.

examples/components/CCFRAUD/preprocessing/spec.yaml
jfomhover: sounds good, applied.

majercakdavid (approved):

ShizeSu06 (approved):
LGTM!

PR #197, opened 2022-12-05
POC for DDP training

majercakdavid:
Proposal for distributed training using PyTorch DDP

CCFRAUD CPU run (this is just to demonstrate that the changes do not break anything): https://ml.azure.com/runs/helpful_yogurt_lqd2q006qb?wsid=/subscriptions/48bbc269-ce89-4f6f-9a12-c6f91fcb772d/resourcegroups/fldev-damajrecak-rg/workspaces/aml-fldevdmdemo0&tid=72f988bf-86f1-41af-91ab-2d7cd011db47#

CCFRAUD GPU run: https://ml.azure.com/runs/frank_pocket_bjn92dmyfd?wsid=/subscriptions/48bbc269-ce89-4f6f-9a12-c6f91fcb772d/resourcegroups/fldev-damajrecak-rg/workspaces/aml-fldevdmdemo0&tid=72f988bf-86f1-41af-91ab-2d7cd011db47#


majercakdavid:
@jfomhover @garg-amit @thomasp-ms @ShizeSu06 would be really great to hear your thoughts and whether we should apply it in the same way to all examples

examples/components/CCFRAUD/traininsilo/run.py
```diff
@@ -457,18 +548,17 @@ def main(cli_args=None):
...
 
     print(f"Running script with arguments: {args}")
-    run(args)
+    print(f"CUDA available: {torch.cuda.is_available()}")
+    cuda_count = torch.cuda.device_count()
+    print(f"Found CUDA devices: {cuda_count}")
 
+    if torch.cuda.device_count() > 1:
+        print("Running DDP training to utilize multiple GPUs.")
+        mp.spawn(run_parallel, args=(cuda_count, args), nprocs=cuda_count, join=True)
```
jfomhover, line 557: let's not use this to run DDP, but instead let's use the AzureML way of creating distributed jobs.

this would greatly simplify the code.

check https://github.com/Azure/azureml-examples/blob/main/tutorials/e2e-distributed-pytorch-image/e2e-object-classification-distributed-pytorch.ipynb
 section "Training a distributed gpu job"

jfomhover: the implementation of the pytorch launcher is taken care of by AzureML, your scripts only has to run the DDP which will consume the env vars (MASTER_ADDR) set up for you by AzureML.

examples/components/CCFRAUD/traininsilo/run.py
```diff
@@ -457,18 +548,17 @@ def main(cli_args=None):
     args = parser.parse_args(cli_args)
 
     print(f"Running script with arguments: {args}")
-    run(args)
+    print(f"CUDA available: {torch.cuda.is_available()}")
```
jfomhover, line 551: let's use logger instead of print

jfomhover (changes requested):
Let's switch to AzureML distribution settings!

examples/components/CCFRAUD/traininsilo/run.py
majercakdavid: I have checked out their solution and I'm not sure about it because either we make pipeline or the component messy. The reason for this is that if you have multi-gpu machine you still need to retrieve all details about it in the pipeline and I'm not really sure how to get the number of GPUs connected to that particular compute. If we make it as a parameter, it's going to be less user friendly.

majercakdavid:
> Let's switch to AzureML distribution settings!

Responded in a thread

examples/components/CCFRAUD/traininsilo/run.py
majercakdavid: @jfomhover what do you think? Can we discuss this maybe in the next meeting?

examples/components/CCFRAUD/traininsilo/run.py
```diff
@@ -420,13 +472,48 @@ def get_arg_parser(parser=None):
...
+        epochs=args.epochs,
+        batch_size=args.batch_size,
+        experiment_name=args.metrics_prefix,
+        iteration_name=args.iteration_name,
+    )
+    trainer.execute(args.checkpoint)
+
 
+def run_parallel(rank, world_size, args):
+    os.environ["MASTER_ADDR"] = "localhost"
```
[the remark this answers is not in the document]
majercakdavid: I'm not 100% convinced - let's discuss.

examples/pipelines/ccfraud/submit.py
```diff
@@ -252,6 +259,28 @@ def fl_ccfraud_basic():
 
         # for each silo, run a distinct training with its own inputs and outputs
         for silo_index, silo_config in enumerate(YAML_CONFIG.federated_learning.silos):
+            # Determine number of processes to deploy on a given compute cluster node
+            ws_compute = ML_CLIENT.compute.get(silo_config.compute)
```
jfomhover, line 263: happy to hear we can obtain this number here, could we put this under a distinct method? (something earlier in the script).

The goal is to enhance readability.

examples/pipelines/ccfraud/submit.py
```diff
@@ -279,6 +308,9 @@ def fl_ccfraud_basic():
             # make sure the compute corresponds to the silo
             silo_training_step.compute = silo_config.compute
 
+            # set distribution according to the number of available GPUs (1 in case of only CPU available)
+            silo_training_step.distribution.process_count_per_instance = silo_processes
```
jfomhover, line 312: could you add a parameter in config.yaml for the number of nodes as well?

examples/pipelines/ccfraud/submit.py
```diff
@@ -122,6 +122,13 @@ def connect_to_aml():
     return ML_CLIENT
 
 
+#############################################
+### GET ML_CLIENT AND COMPUTE INFORMATION ###
+#############################################
+
+ML_CLIENT = connect_to_aml()
```
jfomhover, line 129: IIRC @garg-amit  moved connect_to_aml() below for some reason. I understand this will be problematic for your use case, let's clarify why that was moved.

examples/pipelines/ccfraud/config.yaml
```diff
@@ -49,9 +49,9 @@ training_parameters:
...
   epochs: 3
   lr: 1e-3
-  batch_size: 1000
+  batch_size: 500
   # Property model_name defines model to be used for training and 
   # must be one of the models defined in the traininsilo/model.py 
   # file for the CCFRAUD component. Options include: 
   # SimpleLinear, SimpleLSTM, SimpleVAE
-  model_name: SimpleLSTM
+  model_name: SimpleLinear
```
jfomhover, line 57: no problem for me but.... any reason you're changing the value here?

examples/components/CCFRAUD/traininsilo/run.py
```diff
@@ -112,21 +118,63 @@ def __init__(
...
+                torch.device("cuda", device_id) if torch.cuda.is_available() else "cpu"
+            )
+            if device_id is not None
+            else torch.device("cuda:0" if torch.cuda.is_available() else "cpu")
+        )
+        logger.info(f"Using device: {self.device_}")
+
+        if self._distributed:
+            self._rank = device_id
+            logger.info(f"Rank: {self._rank}")
```
jfomhover, line 134: where is logger defined here?

jfomhover (changes requested):
Thanks for the new version! I'd like to have #nodes as a config as well, and a few comments to address.

examples/components/CCFRAUD/traininsilo/run.py
majercakdavid: line 546

examples/pipelines/ccfraud/config.yaml
majercakdavid: I believe I changed it due to faster execution - just so I can test faster mainly on confidential compute as I was trying to execute it there as well

examples/pipelines/ccfraud/submit.py
garg-amit: That was moved in order to resolve the issue we faced when preserving the precedence hierarchy (PR #209 ). If a user wishes to validate the pipeline without really submitting it, then we don't need to create an ML Client obj.

majercakdavid: Thanks @garg-amit that's reasonable. However, I believe this cannot be applied when searching for specs for compute ahead of time.

garg-amit: Note: We allow users to validate a pipeline without providing rg name, subscription id, etc up front. So, if you wish to create an ML client obj here, it would throw an error if those details are not provided since MLClient obj cannot be created with "None" values.

examples/pipelines/ccfraud/submit.py
```diff
@@ -279,6 +312,15 @@ def fl_ccfraud_basic():
             # make sure the compute corresponds to the silo
             silo_training_step.compute = silo_config.compute
 
+            # set distribution according to the number of available GPUs (1 in case of only CPU available)
```
garg-amit, line 315: How about using an additional flag while submitting a pipeline(maybe --distributed)? n you can put the distributed code in a separate block or fn. In that way, you can keep both the existing as well as the distributed code separate. That would also resolve the MLClient issue.

majercakdavid: Hmm, is it in any way issue that the MLClient is always created?

docs/real-world-examples/ccfraud.md
```diff
@@ -107,3 +107,21 @@ This sample experiment provides multiple models you can try:
 - **SimpleVAE** : a model composed of 2 encoder LSTM layers and 2 decoder LSTM layers that tries to recreate consumed sequence of transactions, the latent space created by encoder is consumed by a linear layer to perform prediction, takes data ordered by time in sequences that overlap each other
 
 To switch between models, please update the `config.yaml` file in `examples/pipelines/ccfraud/`. Look for the field `model_name` in the `training_parameters` section (use `SimpleLinear`, `SimpleLSTM`, or `SimpleVAE`).
+
+## Distributed training
```
jfomhover, line 111: I would not go into the details on how to create a compute here.

Instead, you could say the pipeline is "distributed training ready", and point to the parameters that need to be changed.

+ add a link to the provisioning guide

examples/pipelines/ccfraud/submit.py
```diff
@@ -279,6 +312,15 @@ def fl_ccfraud_basic():
...
+            silo_training_step.distribution.process_count_per_instance = silo_processes
+
+            # set number of instances to distribute training across
+            if hasattr(silo_config, "instance_count"):
```
jfomhover, line 319: this parameter instance_count is not in the config.yaml and is not documented in ccfraud.md

could you add the parameter in config.yaml? (even commented out, with a note that explains why it's relevant) and document it explicitely in your section of ccfraud.md ?

jfomhover (changes requested):
Another set of changes, but mostly for documentation purpose.

jfomhover (approved):

PR #210, opened 2022-12-08
Add data permissions issue to TSG

jfomhover:
Title says it all :)

### DIFF
--- a/docs/tsg.md
+++ b/docs/tsg.md
@@ -63,6 +63,27 @@ During deployment, you may encounter the following exception, even if you cannot
 
 ## Experiment failures
 
+### Issue: Dataset initialization failed DataAccessError(PermissionDenied)
+
+During an experiment, if you encounter the following exception:
+
+```
+Dataset initialization failed: AzureMLException:
+	Message: DataAccessError(PermissionDenied)
+	InnerException None
+	ErrorResponse
+{
+    "error": {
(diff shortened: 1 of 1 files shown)


docs/tsg.md
```diff
@@ -63,6 +64,27 @@ During deployment, you may encounter the following exception, even if you cannot
...
+	InnerException None
+	ErrorResponse
+{
+    "error": {
+        "message": "DataAccessError(PermissionDenied)"
+    }
+}
+```
+
+**Root cause**: This occurs usually in the training phase of our tutorials when the silo compute tries to mount write the orchestrator storage, but doesn't have the permisssions to do so. In some of our tutorials, it happens routinely if, after provisioning a silo or orchestrator, you skipped the section of the tutorial to set the RBAC R/W permissions (example [for open silo](./provisioning/silo_open.md#set-permissions-for-the-silos-compute-to-rw-fromto-the-orchestrator)). This can also happen for other reasons, but what this exception indicates is that the compute's identity doesn't have the required permissions to access/mount the data on the given datastore.
```
garg-amit, line 83: `mount write` or just `mount`?

garg-amit (approved):

docs/tsg.md
jfomhover: fixed

PR #213, opened 2022-12-09
Exploratory Data Analysis (EDA)

majercakdavid:
## Purpose
Demo for demonstration of how one can perform exploratory data analysis in FL

```
[ ] Yes
[ x] No
```

## Pull Request Type
Addition of new demo component

```
[ ] Bugfix
[x ] Feature
[ ] Other... Please describe:
```

## What to Check
Output of `data_analysis` component
* ...


majercakdavid:
Example run: https://ml.azure.com/runs/helpful_peach_y1v89rcfjv?wsid=/subscriptions/48bbc269-ce89-4f6f-9a12-c6f91fcb772d/resourcegroups/fldev-customer1-rg/workspaces/aml-fldemo-customer1&tid=72f988bf-86f1-41af-91ab-2d7cd011db47#

examples/pipelines/ccfraud/config.yaml
```diff
@@ -8,6 +8,8 @@ aml:
 
 # federated learning parameters
 federated_learning:
+  run_data_analysis: true
```
jfomhover, line 11: could you move this parameter under either data_analysis_parameters or training_parameters ?

federated_learning is really a section that is only about configuring the silos and orchestrator.

jfomhover: also, can we turn it off by default? (I think we need to document it better if we want to turn it on by default)

examples/components/utils/data_analysis/run.py
```diff
@@ -0,0 +1,240 @@
...
+    }
+    mlflow.log_dict(
+        column_description, f"{metrics_prefix}/{column_name}/column_description.json"
+    )
+
+    fig, ax = plt.subplots()
+    ax = df[column_name].plot(
+        kind="density", xlim=(df[column_name].min(), df[column_name].max())
+    )
+    mlflow.log_figure(fig, f"{metrics_prefix}/{column_name}/density.png")
```
jfomhover, line 101: can we record all those mlflow calls to the pipeline root level?

This is what is done in the training script, and avoids having users to navigate to each step to find the relevant metrics

examples/components/utils/data_analysis/run.py
```diff
@@ -0,0 +1,240 @@
+import argparse
```
jfomhover, line 1: could we rename this component tabular_data_analysis ? (we might have a text data analysis later?)

jfomhover (changes requested):
For a future PR, I'd love for this data analysis tool to really "gather" something to a central location, so we could plot the data distributions in one single plot.

But the current design is ok. There's a couple comments though about params and others that I'd love to take into account before merging. Thanks!

docs/provisioning/external-silos.md
```diff
@@ -94,8 +94,6 @@ az k8s-extension create --name <extension-name> --extension-type Microsoft.Azure
 ```
 where `<extension-name>` can be chosen arbitrarily.
 
-> Note: if you're using an AKS cluster (as opposed to a local k8s cluster), you'll need to change the `--cluster-type` parameter value from `connectedClusters` to `managedClusters`.
```
jfomhover, line 97: is this intentional here?

majercakdavid:
https://ml.azure.com/runs/sleepy_steelpan_pzpp1l5tqg?wsid=/subscriptions/48bbc269-ce89-4f6f-9a12-c6f91fcb772d/resourcegroups/fldev-damajrecak-rg/workspaces/aml-fldevdmdemo0&tid=72f988bf-86f1-41af-91ab-2d7cd011db47#

examples/components/utils/data_analysis/run.py
majercakdavid: Could we make this change later on when we decide to add additional types of analyses? Maybe we can make it part of this component by just extending it - I believe the code for analysis itself won't be long.

jfomhover (approved):

PR #220, opened 2022-12-12
Generalize aggregate component to Babel

ww-aims:
## Purpose

* Generalize the aggregate component to Babel tasks

Urgency:
- [ ] Low (can wait a week)

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## What to Check
* The aggregate components also work for all the other tasks


examples/components/utils/aggregatemodelweights/run.py
```diff
@@ -39,6 +40,18 @@ def get_arg_parser(parser=None):
     parser.add_argument(
         "--output", type=str, required=True, help="where to write the averaged model"
     )
+    parser.add_argument(
+        "--ancillary_files",
+        type=bool,
+        default=False,
+        help="Wether ancillary files need to be copied",
```
garg-amit, line 47: nit: typo whether

examples/components/utils/aggregatemodelweights/run.py
```diff
@@ -165,7 +176,24 @@ def main(cli_args=None):
     for model_path in model_paths:
         model_handler.add_model(model_path)
 
-    model_handler.save_model(os.path.join(args.output, f"model.{args.extension}"))
+    # check if meta data needed, coppy the whole dir
```
garg-amit, line 179: nit: typo copy

examples/components/utils/aggregatemodelweights/run.py
```diff
@@ -165,7 +176,24 @@ def main(cli_args=None):
...
+    if args.ancillary_files:
+        for src in os.listdir(args.checkpoints[0]):
```
garg-amit, line 181: why are we only saving metadata of  `silo_0`?

examples/components/utils/aggregatemodelweights/run.py
ww-aims: fixed😅

examples/components/utils/aggregatemodelweights/run.py
ww-aims: fixed🤦‍♀️

examples/components/utils/aggregatemodelweights/run.py
```diff
@@ -39,6 +40,18 @@ def get_arg_parser(parser=None):
     parser.add_argument(
         "--output", type=str, required=True, help="where to write the averaged model"
     )
+    parser.add_argument(
+        "--ancillary_files",
+        type=bool,
```
jfomhover, line 45: I think you have to use distutils strtobool here.

examples/components/utils/aggregatemodelweights/spec.yaml
```diff
@@ -44,7 +52,8 @@ command: >-
   $[[${{inputs.input_silo_3}}]] 
   $[[${{inputs.input_silo_4}}]] 
   $[[${{inputs.input_silo_5}}]] 
-
+  --ancillary_files $[[${{inputs.ancillary_files}}]]
```
jfomhover, line 55: you will have to surround `--ancillary_files` as well in the `$[[ ... ]]` block

examples/components/utils/aggregatemodelweights/spec.yaml
```diff
@@ -44,7 +52,8 @@ command: >-
   $[[${{inputs.input_silo_3}}]] 
   $[[${{inputs.input_silo_4}}]] 
   $[[${{inputs.input_silo_5}}]] 
-
+  --ancillary_files $[[${{inputs.ancillary_files}}]] 
+  --out_checkpoint_name $[[${{inputs.out_checkpoint_name}}]]
```
jfomhover, line 56: same here, the `--out_checkpoint_name` has to be within the optional block `$[[..]]`

examples/components/utils/aggregatemodelweights/run.py
garg-amit: I think the tokenizer files and even hyperparams could differ for each silo. Wouldn't that be the case? As silo local trainings are independent of each other. (Note: It might not be the case in your current scenario but if we want to generalize it then this is also possible, right?)

examples/components/utils/aggregatemodelweights/spec.yaml
ww-aims: that's right, good catch!

examples/components/utils/aggregatemodelweights/run.py
ww-aims: yep! Absolutely right

examples/components/utils/aggregatemodelweights/run.py
ww-aims: For the tokenizer files, here is an example:
{
  "bos_token": "[CLS]",
  "cls_token": "[CLS]",
  "eos_token": "[SEP]",
  "mask_token": "[MASK]",
  "pad_token": "[PAD]",
  "sep_token": "[SEP]",
  "unk_token": "[UNK]"
}
I think as long as these specifications are consistent for the datasets used in different silos, they should be the same. For the hyper parameter files, I agree they could be different for each silo. I will change the component to accommodate that! Thanks!

jfomhover: I think your point Amit is very valid, we might have different metadata in different silos. Although it might be a minority of use cases. Would it be ok to merge the PR with its current design (once other comments are resolved) and investigate a more flexible design in another PR?

garg-amit: Sounds good to me! We can address this in another PR

garg-amit (approved):

jfomhover (approved):

PR #237, opened 2023-01-18
Differential Privacy

garg-amit:
## Purpose

* Implement differential privacy across all examples.

Urgency:
- [ ] Low (can wait a week)

```
[x] Yes
[ ] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

Submit any pipeline by enabling differential privacy(update config file) and wait for its completion.


examples/pipelines/ccfraud/config.yaml
```diff
@@ -63,3 +63,12 @@ training_parameters:
   # file for the CCFRAUD component. Options include: 
   # SimpleLinear, SimpleLSTM, SimpleVAE
   model_name: SimpleLinear
+
+  # Differential privacy
+  dp: true # Flag to enable/disable differential privacy
+  dp_target_epsilon: 50.0 # Smaller epsilon means more privacy, more noise
```
majercakdavid, line 69: Maybe add suggested ranges for epsilon to the comment?

majercakdavid (commented):
Can you add successful run for at least one of the examples?

examples/pipelines/ccfraud/config.yaml
garg-amit: Added more info

garg-amit:
> Can you add successful run for at least one of the examples?

Shared with you offline

jfomhover (approved):
Looks good!

examples/components/MNIST/traininsilo/run.py
```diff
@@ -229,7 +229,7 @@ def local_train(self, checkpoint):
             test_acc = 0.0
 
             for epoch in range(self._epochs):
-
+                self.model_.train()
```
ShizeSu06, line 232: why there is a duplicated self.model_.train() here?
image

garg-amit: nice catch! fixed

examples/components/HELLOWORLD/traininsilo/run.py
```diff
@@ -4,6 +4,7 @@
 import sys
 import glob
 import os
+from distutils.util import strtobool
```
ShizeSu06, line 7: The helloworld example doesn't include any meaningful model training, and it is also irrelevant for dp training (we can see we don't even import opacus in this run script). In this case, would that be slightly better to just drop all the changes about dp in this hello world example? Are there any other potential issues it would cause if we do that?

ShizeSu06 (approved):

examples/components/HELLOWORLD/traininsilo/run.py
garg-amit: Agree! that doesn't make sense to implement DP in the HELLOWORLD example but the `submit.py` is shared by both MNIST and HELLOWORLD example so we have to do the respective changes in the HELLOWORLD example as well else have to put extra `if else` conditions in the submit.py

ShizeSu06: got it, sounds good to me.

PR #246, opened 2023-01-23
Differential Privacy - MNIST

garg-amit:
## Purpose

* Implement differential privacy in the MNIST example.
* Adding noise as well as clipping gradient is taken care of by the Opacus package itself.

Urgency:
- [ ] Low (can wait a week)

```
[x] Yes
[ ] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

Submit the MNIST example by enabling differential privacy and see if there are any errors.


thomasp-ms (changes requested):
I submitted 2 jobs (one with DP, one without). A couple meta comments. Blocking because of the last one.

- shouldn't we do this for factory code as well? (not saying necessarily in this PR)
- is it explained anywhere in the repo what DP is, and that it can be turned on and off? and also a brief explanation about what parameters do, in addition to their name? (not saying necessarily in this PR)
- we should have caught this earlier, but right now I believe the quickstart is broken. By default, it creates just one compute per silo (compute2 = false by default in the bicep template) and the config file here references the second compute. This needs to be addressed before we release. (If you want to address in a distinct PR, please let me know and I'll just approve this one.)

examples/components/MNIST/traininsilo/run.py
```diff
@@ -32,6 +40,9 @@ def __init__(
             lr (float, optional): Learning rate. Defaults to 0.01
             epochs (int, optional): Epochs. Defaults to 1
             batch_size (int, optional): DataLoader batch size. Defaults to 64
+            dp (bool, optional): Differential Privacy
```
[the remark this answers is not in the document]
garg-amit: done

examples/components/MNIST/traininsilo/run.py
```diff
@@ -171,7 +205,7 @@ def local_train(self, checkpoint):
 
                 running_loss = 0.0
                 num_of_batches_before_logging = 100
-
+                self.model_.train()
```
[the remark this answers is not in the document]
garg-amit: I notice that we are calling `self.model_.eval()` at the end of each epoch. In order to ensure that the model is in the training state at the start of each epoch, I moved the `self.model_.train()` at the beginning of the `for-loop`.

garg-amit:
Thank you for reviewing, @thomasp-ms!
[1] Agree! This is coming in the upcoming [PR](https://github.com/Azure-Samples/azure-ml-federated-learning/pull/237).
[2][3] I've updated the documentation as well as added a description of each param in the config file.

ShizeSu06 (changes requested):
Thanks Amit for preparing the PR! The implementation itself looks correct to me ovrall. Some meta comments for improvements:
[1] I agree with Thomas that we should prepare some documentation/readme about DP concept/technique as well as key tips/pitfalls, which could be in separate PR. For this release, we won't target for a lengthy comprehensive one, and a brief concise intro/summary might be  sufficient. I'd help preparing a draft on this and share/collaborate with Amit.
[2] It might be good to add code for check incompatible layers and model validation/fix, since it is not uncommon to have some incompatible layers (and it is simple to add those codes). I.e., sth like below:
image
image

[3] The dp example here is using "noise multiplier" when specifying the privacy engine, which is fine and simple in implementation. But people might not have a good intuitive idea about what "noise multiplier = 1" means (e.g., is that stand for a typical noise level, large noise level, or small noise level, etc.). DP framework is most well-known for the epsilon-delta mathematical schema where delta might default to sth like min (a tiny number, 1/N (here N is number of data samples)), and there are some kind of agreements about acceptable level of privacy budget epsilon in industry (e.g., epsilon=4 is a typical accepted threshold in many applications). Note that, noise multiplier, epsilon and delta are not independent from each other, and given two of them we could compute the third one. In such examples where we use "noise multiplier" for privacy engine, it might be helpful to also log/output the consumed privacy budget epsilon (using sth like "epsilon = privacy_engine.get_epsilon(DELTA)") to inform users such key privacy info. 
[4] In the doc or examples, it might be good to also share other standard way of DP implementation, e.g., dp training using target_epsilon and target_delta input in privacy_engine (rather than noise multipler). I.e., sth like below: 
image

[5] There are some questions about correctly compute privacy budget epsilon across different FL iterations for which it requires a bit more discussion, and will discuss offline.
[6] In my offline chat with Jeff, he asked whether this DP solution could work for distributed scenario. I'd expect this solution to also work naturally for distributed scenario, but we should try running testing experiments to verify before the official release.

examples/components/MNIST/traininsilo/run.py
```diff
@@ -77,6 +88,28 @@ def __init__(
...
+        # DP
+        logger.info(f"DP: {dp}")
+        if dp:
+            if not ModuleValidator.is_valid(self.model_):
+                self.model_ = ModuleValidator.fix(self.model_)
+
+        self.optimizer_ = SGD(self.model_.parameters(), lr=lr, momentum=0.9)
+
+        if dp:
+            privacy_engine = PrivacyEngine(secure_mode=False)
```
ShizeSu06, line 100: nit: might be good to add some doc strings to explain "secure_mode" and point out potential pitfall about it. Note that, using secure_mode=False is better for performance (i.e., faster) but it has known weakness/vulnerability that doesn't meet the safety bar for scenarios targeting for shipping model to production. So it is good for faster exploration/experimenting, but need to turn on (i.e., secure_model=True) when training DP model to ship to production. Btw, below I shared official docstring explanation about secure_mode:

secure_mode: Set to ``True`` if cryptographically strong DP guarantee is
                required. ``secure_mode=True`` uses secure random number generator for
                noise and shuffling (as opposed to pseudo-rng in vanilla PyTorch) and
                prevents certain floating-point arithmetic-based attacks.
                See :meth:`~opacus.optimizers.optimizer._generate_noise` for details.
                When set to ``True`` requires ``torchcsprng`` to be installed

garg-amit: makes sense! done

garg-amit:
Thank you for reviewing, @ShizeSu06!
As discussed, I've changed the approach to use `make_private_with_epsilon` method by default.

examples/components/MNIST/traininsilo/run.py
```diff
@@ -77,6 +92,48 @@ def __init__(
...
+                self.model_,
+                self.optimizer_,
+                self.train_loader_,
+            ) = privacy_engine.make_private_with_epsilon(
+                module=self.model_,
+                optimizer=self.optimizer_,
+                data_loader=self.train_loader_,
+                epochs=dp_epochs,
+                target_epsilon=dp_target_epsilon,
+                target_delta=dp_target_delta,
```
ShizeSu06, line 121: Minor suggestion: it might be helpful to log/print out the corresponding "noise multiplier" for such private_engine with the targeted epsilon and delta. You could get that "noise multiplier" value by optimizer.noise_multiplier after above definition of private_engine. It also helps explain a bit that epsilon, delta and noise multiplier are not independent from each other (we could compute the 3rd one given the other two).

examples/components/MNIST/traininsilo/spec.yaml
```diff
@@ -43,6 +43,31 @@ inputs:
     description: batch size
     default: 64
     optional: true
+  dp:
+    type: boolean
+    description: differential privacy
+    default: false
+    optional: true
+  dp_epochs:
```
ShizeSu06, line 51: @garg-amit dp_epoch, epoch (for a local train), and num_of_iterations are not independent from each other. dp_epoch=epoch* num_of_iterations. So in order to avoid potential user mistakes that they arbitrarily specify the dp_epoch, epoch and num_of_iteratios which doesn't satisfy the required relationship of dp_epoch=epoch* num_of_iterations, I suggest that we remove this dp_epoch input arg (and thus only have epoch and num_of_iterations inut arg), and compute dp_epoch using the relationship dp_epoch=epoch* num_of_iterations.

examples/components/MNIST/traininsilo/run.py
garg-amit: agree! done

examples/components/MNIST/traininsilo/spec.yaml
garg-amit: makes sense! Unfortunately, the input parameter "num of iterations" is not available at this time (the "iteration num" field only specifies the current iteration number). I still believe it would be better to use `num of iteration` as an input parameter rather than `dp epochs` to prevent misunderstandings. I'll do the required changes

garg-amit: done!

examples/components/MNIST/traininsilo/spec.yaml
```diff
@@ -24,6 +24,11 @@ inputs:
     description: Iteration number
     default: 1
     optional: true
+  total_num_of_iterations:
```
ShizeSu06, line 27: Nit: since "total_num_of_iterations" is only required for DP purpose, it might be slightly better to put this input in the same section as other dp-related args, e.g., after line 55 or line 70, and also add a doc string clarifying that those arg inputs (dp, dp-target_epsilon, dp_target_delta, dp_max_grad_norm, and total_num_of_iterations are defined for the only purpose of DP and can be ignored when users don't want to use DP.

examples/pipelines/fl_cross_silo_literal/submit.py
```diff
@@ -282,10 +282,20 @@ def fl_cross_silo_internal_basic():
...
+                # DP target delta
+                dp_target_delta=YAML_CONFIG.training_parameters.dp_target_delta,
+                # DP max gradient norm
+                dp_max_grad_norm=YAML_CONFIG.training_parameters.dp_max_grad_norm,
                 # Silo name/identifier
                 metrics_prefix=silo_config.name,
                 # Iteration number
                 iteration_num=iteration,
+                # Total num of iterations
+                total_num_of_iterations=YAML_CONFIG.training_parameters.num_of_iterations,
```
ShizeSu06, line 298: Nit; similarly, since "total_num_of_iterations" is only required for DP purpose, it might be slightly better to put this input arg together with those other DP-related parameters (line 285-292), and added brief docstring mention this arg is only required for DP.

ShizeSu06: @garg-amit it seems you forgot to update this in submit.py in your last commit.

examples/components/MNIST/traininsilo/run.py
```diff
@@ -32,6 +42,11 @@ def __init__(
             lr (float, optional): Learning rate. Defaults to 0.01
             epochs (int, optional): Epochs. Defaults to 1
             batch_size (int, optional): DataLoader batch size. Defaults to 64
+            dp (bool, optional): Differential Privacy. Default is False (Note: dp, dp_target_epsilon, dp_target_delta, dp_max_grad_norm, and total_num_of_iterations are defined for the only purpose of DP and can be ignored when users don't want to use Differential Privacy)
```
ShizeSu06, line 45: For "Default is False", I saw capital "False" here, but in the spec.yaml file, config.yaml file as well as quickstart.md, I saw lower case "false/true" there. Better to verify and also be consistent to avoid any potential pitfalls. (generally Boolean values need to use capitalized "True/False" but for special contexts "true/false" might also possibly work fine, but better to verify and be consistent.)

garg-amit: I intentionally did that to be consistent with other files bcoz in all our yaml/md file, the boolean is defined with lowercase t/f ..For instance, [azure-ml-federated-learning/spec.yaml at release-05 · Azure-Samples/azure-ml-federated-learning (github.com)](https://github.com/Azure-Samples/azure-ml-federated-learning/blob/release-05/examples/components/HELLOWORLD/traininsilo/spec.yaml#L21).

examples/components/MNIST/traininsilo/run.py
```diff
@@ -289,7 +350,12 @@ def get_arg_parser(parser=None):
     parser.add_argument(
         "--iteration_num", type=int, required=False, help="Iteration number"
     )
-
+    parser.add_argument(
```
ShizeSu06, line 353: nit: similarly, might be slightly better to move this "total_num_of_iterations" to those dp args section, since this arg is only for DP purpose.

jfomhover (approved):

PR #267, opened 2023-02-10
Fix CUDA and DDP training

majercakdavid:
## Purpose
Update package used for cuda runtime for PyTorch according to installation instructions. Provide fixes for DDP training for some samples

Urgency:
- [X] High (needs review today)
- [ ] Low (can wait a week)

```
[X] Yes
[ ] No
```

## Pull Request Type

```
[X] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...

### DIFF
--- a/examples/components/CCFRAUD/traininsilo/run.py
+++ b/examples/components/CCFRAUD/traininsilo/run.py
@@ -598,7 +598,7 @@ def run(args):
     )
     trainer.execute(args.checkpoint)
 
-    if torch.cuda.is_available() or int(os.environ["WORLD_SIZE"]) > 1:
+    if int(os.environ["WORLD_SIZE"]) > 1:
         dist.destroy_process_group()
 
 
--- a/examples/components/CCFRAUD/traininsilo/run.py
+++ b/examples/components/CCFRAUD/traininsilo/run.py
@@ -594,7 +594,8 @@ def run(args):
         experiment_name=args.metrics_prefix,
         iteration_name=args.iteration_name,
         device_id=int(os.environ["RANK"]),
-        distributed=int(os.environ.get("WORLD_SIZE", "1")) > 1 and torch.cuda.is_available(),
+        distributed=int(os.environ.get("WORLD_SIZE", "1")) > 1
+        and torch.cuda.is_available(),
     )
     trainer.execute(args.checkpoint)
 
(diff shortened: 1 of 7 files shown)


examples/components/CCFRAUD/traininsilo/run.py
```diff
@@ -598,7 +598,7 @@ def run(args):
     )
     trainer.execute(args.checkpoint)
 
-    if torch.cuda.is_available() or int(os.environ["WORLD_SIZE"]) > 1:
```
garg-amit, line 601: let's handle a case if there's no `WORLD_SIZE` env var and apply the changes across all examples.

jfomhover: yep could we use os.environ.get("WORLD_SIZE", 1) so it would work even if the env var is not set?

jfomhover: (I actually don't remember if the variable is set when the world size is actually just 1 and the job is not distributed)

examples/components/NER/traininsilo/run.py
```diff
@@ -180,13 +180,13 @@ def __init__(
         logger.info(f"Trainable parameters: {trainable_params}")
 
         self.model_.train()
+        self.model_.to(self.device_)
```
jfomhover, line 183: Is .train() interfering with .to() somehow? I wonder if it's more proper to first do .to() then only .train()

examples/components/NER/traininsilo/run.py
```diff
@@ -180,13 +180,13 @@ def __init__(
...
         if self._distributed:
             self.model_ = DDP(
```
jfomhover, line 185: You're moving the call to .to(device) much earlier than before, is there a reason for it?

In particular, when you instanciate DDP(), I don't know how that works with .to(), is it being taken care of internally in DDP()?

jfomhover: forget about this comment, no pb here

examples/components/NER/traininsilo/run.py
majercakdavid: Should not matter according to documentation: https://pytorch.org/docs/stable/_modules/torch/nn/modules/module.html#Module.train 

The only reason why this was moved here from line 189 is that DDP is conditioned under having model in GPU.

garg-amit: If this is not required for DDP, we can move this to line 194 as a model has to be in the `training mode` when applying changes for `differential-privacy`.(By default, huggingface loads a model in the `eval mode`)

garg-amit (approved):

garg-amit (approved):
let's update arm templates in a separate PR

jfomhover (approved):

PR #268, opened 2023-02-13
FL native contract backed by the factory engine

garg-amit:
## Purpose

* Replaced the factory code's `submit.py` with the actual FL native contract.
* Aligned the `config.yaml` file to the spec CLI YAML file.
* The factory engine backs the pipeline.

Urgency:
- [ ] Low (can wait a week)

```
[x] Yes
[ ] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

Submit the factory pipeline and see if everything is as expected.

MNIST run: https://ml.azure.com/experiments/id/cd48cb1d-2cee-4b10-8a76-2d094b2a5868/runs/patient_kite_n53jmp1kv6?wsid=/subscriptions/48bbc269-ce89-4f6f-9a12-c6f91fcb772d/resourcegroups/amitgarg-fldev-rg/workspaces/aml-testfactory&tid=72f988bf-86f1-41af-91ab-2d7cd011db47#


examples/pipelines/fl_cross_silo_factory/submit.py
```diff
@@ -282,132 +318,58 @@ def silo_scatter_subgraph(
...
-    log_format = logging.Formatter("[%(levelname)s] - %(message)s")
-    handler = logging.StreamHandler(sys.stdout)
-    handler.setLevel(logging.DEBUG)
-    handler.setFormatter(log_format)
-    logger.addHandler(handler)
-
-pipeline_job = builder.build_flexible_fl_pipeline(
-    # building the FL graph requires two arguments that can be either a component or a pipeline
-    # scatter is our subgraph above
+pipeline_job, fl_factory = scatter_gather(
```
jfomhover, line 355: could we split the factory in two: 1) the scatter_gather() engine and 2) the soft validation engine, which could be in a completely distinct class (and import).

The goal would be to get the scatter_gather() helper very cleanly defined into a python script of its own, and the soft_validation as a completely distinct source code, that we can choose to pull in or not.

examples/pipelines/fl_cross_silo_factory/submit.py
```diff
@@ -282,132 +318,58 @@ def silo_scatter_subgraph(
...
-            mode=silo_config.training_data.mode,
-            path=silo_config.training_data.path,
-        ),
-        raw_test_data=Input(
-            type=silo_config.testing_data.type,
-            mode=silo_config.testing_data.mode,
-            path=silo_config.testing_data.path,
-        ),
-    )
+scatter_configs = get_scatter_configs()
```
jfomhover, line 351: I would actually argue it would be better to have list comprehension here instead of wrapping this up inside a function.

This way, the FL contract is laid out clearly in this one section of the code, and developers don't have to check your function up above this script to know how to write silo configs.

a simple list comprehension could do the trick

examples/pipelines/fl_cross_silo_factory/submit.py
```diff
@@ -282,132 +318,58 @@ def silo_scatter_subgraph(
...
+gather_config = get_gather_config()
```
jfomhover, line 352: same here, I'd unpack this function here, not wrap up the gather config parsing in a function.

examples/pipelines/fl_cross_silo_factory/submit.py
garg-amit: Agree! done

examples/pipelines/fl_cross_silo_factory/submit.py
garg-amit: done! thanks for the code snippet, Jeff!

examples/pipelines/fl_cross_silo_factory/submit.py
```diff
@@ -42,9 +42,6 @@
 # to handle yaml config easily
```
jfomhover, line 42: Should we remove the term "factory" from this submit.py entirely? (and in our docs as well?)

examples/pipelines/fl_cross_silo_factory/submit.py
```diff
@@ -190,13 +180,33 @@ def connect_to_aml():
...
-### C. CUSTOM DS CODE ###
+### C. Scatter/Gather pipelines ###
 #########################
 
 # This is your section, please modify anything here following the guidelines
 # in the docstrings.
 
 
+@pipeline
+def gather_pipeline(
```
jfomhover, line 191: should we just use aggregate_component directly?

examples/pipelines/fl_cross_silo_factory/submit.py
```diff
@@ -282,132 +286,71 @@ def silo_scatter_subgraph(
...
+
+scatter_configs = [
+    {
+        "inputs": {
+            "raw_train_data": Input(**dict(silo_config["inputs"])["raw_training_data"]),
+            "raw_test_data": Input(**dict(silo_config["inputs"])["raw_testing_data"]),
+        },
+        "computes": silo_config["computes"],
+        "datastore": silo_config["datastore"],
+        "name": silo_config["name"],
```
jfomhover, line 322: I don't think "name" is part of our contract? Should we just inject "name" under "inputs" as a string?

examples/pipelines/fl_cross_silo_factory/submit.py
```diff
@@ -282,132 +286,71 @@ def silo_scatter_subgraph(
...
-    pipeline_job,
-    raise_exception=not (
-        args.ignore_validation
-    ),  # set to False if you know what you're doing
-)
+if not args.ignore_validation:
+    from fl_helper import FLValidationEngine
 
+    fl_val_engine = FLValidationEngine(scatter_configs, gather_config)
+    fl_val_engine.soft_validate(pipeline_job)
```
jfomhover, line 349: instead of running soft_validate() only if not ignore_validation, we could use soft_validate's argument raise_exception=True/False to ALWAYS do the validation, but not except if ignored

garg-amit: done

examples/pipelines/fl_cross_silo_factory/submit.py
garg-amit: done

examples/pipelines/fl_cross_silo_factory/submit.py
garg-amit: done

examples/pipelines/fl_cross_silo_factory/submit.py
garg-amit: done! removed from everywhere

jfomhover (approved):

PR #274, opened 2023-02-22
Fix DDP accuracy and add batch_size parameter for pneumonia

ww-aims:
## Purpose
Fix test accuracy for DDP setting, and add batch size as a parameter to the config 

Urgency:
- Medium (needs review within a few days - most common)
- [ ] Low (can wait a week)

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```


garg-amit (approved):

examples/components/PNEUMONIA/traininsilo/run.py
```diff
@@ -353,16 +355,18 @@ def test(self):
...
                 data, target = data.to(self.device_), target.to(self.device_)
+                num_sample += data.shape[0]
                 output = self.model_(data)
                 test_loss += self.loss_(output, target).item()
                 pred = output.argmax(dim=1, keepdim=True)
                 correct += pred.eq(target.view_as(pred)).sum().item()
 
-        test_loss /= len(self.test_loader_.dataset)
-        acc = correct / len(self.test_loader_.dataset)
+        test_loss /= num_sample
```
majercakdavid, line 368: afaik, by default the loss functions in PyTorch calculate mean loss across samples in batch, thus I believe correct way to calculate average loss is to divide this by number of batches

examples/components/PNEUMONIA/traininsilo/run.py
```diff
@@ -430,6 +434,7 @@ def get_arg_parser(parser=None):
         required=False,
         help="Total number of epochs for local training.",
     )
+    parser.add_argument("--batch_size", type=int, required=False, help="Batch Size")
```
jfomhover, line 437: you will have to set default=32 here, or else I believe you'll use None as default value, and your code might fail line 134 when None is given as value for batch_size to DataLoader

jfomhover (changes requested):

examples/components/PNEUMONIA/traininsilo/run.py
ww-aims: Absolutely, that makes sense. Just one thing to confirm -- it looks like the "batch_size" as we implemented in all examples is actually a required parameter, otherwise we will get key error in the submit.py saying that the "batch_size" does not exist in the config. Please correct me if that is wrong.🙂

jfomhover: yes it's required, so somehow this required=True here is redundant, but if at some point we change the definition somewhere else, I don't want this script to fail. Thx for fixing it!

jfomhover (approved):

PR #303, opened 2023-03-19
VAE and single-shot transfer for faster VFL training

majercakdavid:
## Purpose
Showcase possible utilization of Variational Autoencoders for faster VFL trainings requiring less communication.

- **Base case CCFRAUD_VERTICAL:**
  - **Upload data:** https://ml.azure.com/runs/jolly_basin_jxqw60b8zl?wsid=/subscriptions/48bbc269-ce89-4f6f-9a12-c6f91fcb772d/resourcegroups/vertical-fl-sandbox-rg/workspaces/aml-vfl&tid=72f988bf-86f1-41af-91ab-2d7cd011db47#
  - **Train:** https://ml.azure.com/runs/patient_feast_jxc7nq7c51?wsid=/subscriptions/48bbc269-ce89-4f6f-9a12-c6f91fcb772d/resourcegroups/vertical-fl-sandbox-rg/workspaces/aml-vfl&tid=72f988bf-86f1-41af-91ab-2d7cd011db47#
- **FedOnce:**
  - **Upload data:** https://ml.azure.com/runs/honest_double_7htyd26x84?wsid=/subscriptions/48bbc269-ce89-4f6f-9a12-c6f91fcb772d/resourcegroups/vertical-fl-sandbox-rg/workspaces/aml-vfl&tid=72f988bf-86f1-41af-91ab-2d7cd011db47#
  - **Train:** https://ml.azure.com/runs/stoic_cartoon_nt19k24zhn?wsid=/subscriptions/48bbc269-ce89-4f6f-9a12-c6f91fcb772d/resourcegroups/vertical-fl-sandbox-rg/workspaces/aml-vfl&tid=72f988bf-86f1-41af-91ab-2d7cd011db47#

Urgency:
- [x] High (needs review today)
- [ ] Low (can wait a week)

```
[X] Yes
[ ] No
```

## Pull Request Type

```
[ ] Bugfix
[X] Feature
[ ] Other... Please describe:
```


garg-amit (commented):
It's worth considering adding a new parameter “pre-train”(maybe) to the training script itself instead of having a new pretraining component, which would significantly reduce code duplication. As discussed, the major difference would be a model, data loader, and a training loop (coz of the communication layer) which can be easily implemented by introducing either a new method or a new class in the contributor.py file. Let's wait for others' opinions.

examples/pipelines/ccfraud_vertical/submit.py
```diff
@@ -248,11 +252,84 @@ def fl_ccfraud_vertical_basic():
...
+
+    ###############################
+    ### (OPTIONAL) PRE-TRAINING ###
+    ###############################
+
+    embeddings_paths = {}
+
+    if YAML_CONFIG.training_parameters.model_name == "SimpleVAE":
+        # for each silo, run a distinct training with its own inputs and outputs
+        for silo_index, silo_config in enumerate(
```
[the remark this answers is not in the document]
majercakdavid: The reason I picked this way is that I did not like to write everywhere `silo_index+1` so I opted in for just skipping first element.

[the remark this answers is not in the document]
garg-amit: sorry, but why do you want to use `silo_index+1` everywhere as it already starts with index values as 1?

majercakdavid:
> It's worth considering adding a new parameter “pre-train”(maybe) to the training script itself instead of having a new pretraining component, which would significantly reduce code duplication. As discussed, the major difference would be a model, data loader, and a training loop (coz of the communication layer) which can be easily implemented by introducing either a new method or a new class in the contributor.py file. Let's wait for others' opinions.

Thank you very much for invaluable input @garg-amit. I have considered this approach when originally developing the solution, however, let me state the reason why I have chosen the way it is written. 

First and foremost, I like the distinction between the components in the pipeline. If we want to have an idiomatic solution that user easily understand it is probably preferable to avoid creating components that do everything. The main point of this repository is educational and not production solution that has steep understanding and learning curve.

Other than that, the pretraining component has following differences when compared to the training component:

- Different inputs/output (here I would like to stress around additional output for embeddings which could be misleading in default case)
- Different dataloader/sampler because pretraining does not require alignment across the samples
- No communication layer because pretraining is completely independent of samples on other nodes and can utilize non-overlapping samples
- Different model due to the nature of the problem. Regular VFL requires classifier whereas pretraining requires architecture aimed for unsupervised learning.
- Different training loop due to the absence of communication layer/overhead and if we would merge both components into one, we would have either two training files, which would increase complexity of the solution, or have two different functions for training, which would make it inconsistent in regard to other samples.

jfomhover:
Looking at it, please consider me as required before merging.

docs/tutorials/vertical-fl.md
```diff
@@ -67,7 +67,7 @@ This can all be performed with ease using a data provisioning pipeline. To run i
...
+It is an ongoing research topic on how the model can be orchestrated in VFL. We have decided to make the most common approach as default by splitting it between the host and contributors, also referred to as **split learning**, this approach can be easily altered by moving layers between parties to hosting whole model on contributors while host provides only aggregation and/or activation function. We believe that this can better demonstrate capabilities of VFL on AzureML and most of the existing models can be easily split without requiring too much work. However, **CCFRAUD** example is also runnable with unsupervised latent representation learning through Variational Autoencoder (VAE), which is shared once with the host and thus significantly reduces communication overhead. The host then uses these latent representations to predict the target values. More about this approach, and how to set it up, can be read in the section [below](#single-shot-learning-through-variational-autoencoder).
```
jfomhover, line 70: For educational value, I'm beginning to think we might need to split this into two distinct pipelines?

jfomhover: The tutorial itself starts to become something very configurable, maybe too much.

examples/components/CCFRAUD_VERTICAL/traininsilo/host_spec.yaml
```diff
@@ -28,6 +28,18 @@ inputs:
     type: uri_folder
     description: a given pre-existing checkpoint
     optional: true
+  contributor_1_embeddings:
```
jfomhover, line 31: I'm not a fan of the idea of having a fixed number of inputs, which depends on the number of silos. Is there any other way ?

Could we do a first scatter/gather phase just to generate embeddings in an aggregated folder, then do the training in silos?

jfomhover:
I've filed an issue #328 that I'd like you to implement first, in order to see more clearly each of the VFL samples. I believe VAE could be a distinct concept within the vertical-fl page, and a distinct tutorial maybe (TBD).

majercakdavid:
> Looking at it, please consider me as required before merging.
> 
> Could you provide a link to a paper referencing this approach? I'd like to understand how canonical this approach is.

https://arxiv.org/pdf/2208.10278.pdf

docs/tutorials/vertical-fl.md
majercakdavid: We can definitely open this topic during the deep dive tomorrow. The only change in the pipeline itself, as it goes for the user, is the name of the network being trained `SimplerLinear` -> `SimpleVAE`, so the **number of configurations does not change**. The downside I see here is that the approach is a bit hidden and having separate pipeline may help it to stand out?

examples/components/CCFRAUD_VERTICAL/traininsilo/host_spec.yaml
majercakdavid: This is the same issue as with `aggregatemodelweights` component and I believe we have not figured out any other better way for now, right?

majercakdavid: Neither me but I believe we do no have better solution, otherwise we would already implement it for `aggregatemodelweights` right?

examples/components/CCFRAUD_VERTICAL/traininsilo/conda.yaml
```diff
@@ -15,3 +15,4 @@ dependencies:
     - pandas==1.5.2
     - torchmetrics==0.10.3
     - redis==4.5.1
+    - tqdm
```
jfomhover, line 18: please remove this line, looks like you're not using it in either contributor or host anyway (and it would need to pin down a specific version)

examples/components/CCFRAUD_VERTICAL/traininsilo/host.py
```diff
@@ -11,6 +11,7 @@
 
 import mlflow
 import torch
+from tqdm import tqdm
```
jfomhover, line 14: unused import?

examples/components/CCFRAUD_VERTICAL/traininsilo/host.py
```diff
@@ -498,8 +495,11 @@ def run(args, global_comm):
         args (argparse.namespace): command line arguments provided to script
     """
 
+    os.environ["CUDA_DEVICE_ORDER"] = "PCI_BUS_ID"
```
jfomhover, line 498: can you clarify in a comment why we're doing this here?

examples/components/CCFRAUD_VERTICAL_FEDONCE/pretraining/conda.yaml
```diff
@@ -0,0 +1,18 @@
...
+  - defaults
+  - pytorch
+  - nvidia
+dependencies:
+  - python=3.8
+  - pip=22.3.1
+  - pytorch=1.13.1
+  - pytorch-cuda=11.6
+  - pip:
+    - azureml-core==1.47.0
```
jfomhover, line 12: can you reduce dependencies as needed? I believe those are dependencies of the aml_comm helper

examples/components/CCFRAUD_VERTICAL_FEDONCE/pretraining/spec.yaml
```diff
@@ -0,0 +1,73 @@
...
+    type: integer
+    description: total number of epochs for local training
+    default: 10
+    optional: true
+  batch_size:
+    type: integer
+    description: batch size
+    default: 100
+    optional: true
+  global_size:
```
jfomhover, line 41: Those args should not be necessary here?

examples/components/CCFRAUD_VERTICAL_FEDONCE/pretraining/run.py
```diff
@@ -0,0 +1,485 @@
...
+        return {
+            f"{self._prefix}_{name}": value / self._batch_count_global
+            for name, value in self._running_global_metrics.items()
+        }
+
+
+class CCFraudTrainer:
+    def __init__(
+        self,
+        global_rank,
```
jfomhover, line 79: if I understand things correctly, I don't think global_rank is a required parameter in this script?

examples/components/CCFRAUD_VERTICAL_FEDONCE/traininsilo/run.py
```diff
@@ -0,0 +1,530 @@
...
+        return {
+            f"{self._prefix}_{name}": value / self._batch_count_global
+            for name, value in self._running_global_metrics.items()
+        }
+
+
+class CCFraudTrainer:
+    def __init__(
+        self,
+        global_rank,
```
jfomhover, line 80: same here, remove those unused params

examples/components/CCFRAUD_VERTICAL_FEDONCE/traininsilo/spec.yaml
```diff
@@ -0,0 +1,99 @@
...
+    type: integer
+    description: total number of epochs for local training
+    default: 10
+    optional: true
+  batch_size:
+    type: integer
+    description: batch size
+    default: 100
+    optional: true
+  global_size:
```
jfomhover, line 53: note to remove unused params

examples/pipelines/ccfraud_vertical_fedonce/submit.py
```diff
@@ -0,0 +1,412 @@
+"""Federated Learning Cross-Silo Vertical basic pipeline for ccfraud.
```
jfomhover, line 1: the pipeline will fail if the number of silos is above 3, can we either 1) issue an exception if the number of silos is not 3 during this submit.py or 2) make the number of embeddings in traininsilo flexible with a maximum number (ex: 5) and optional inputs?

jfomhover (changes requested):
Thanks overall the comments I have are related to unused params, and some warning for users about number of silos in submit.py

examples/components/CCFRAUD_VERTICAL_FEDONCE/pretraining/run.py
majercakdavid: this is for logging purposes (so we are able to distinguish source of the logs on the pipeline level)

examples/components/CCFRAUD_VERTICAL_FEDONCE/traininsilo/run.py
majercakdavid: just the same reason as previously

examples/components/CCFRAUD_VERTICAL_FEDONCE/pretraining/run.py
jfomhover: then you should use something similar to what was done for the horizontal FL cases, also.... I don't see where `global_rank` is used at all in the code so I'm not sure how you're consuming it as a logging purpose?

jfomhover: (in horizontal FL, we use metrics_prefix for that)

majercakdavid: image

jfomhover (approved):

PR #304, opened 2023-03-20
Provide fixes and unified approach for measuring time across communication techniques in VFL

majercakdavid:
## Purpose
More precise and fine grained metrics for communication

Urgency:
- [X] Medium (needs review within a few days - most common)
- [ ] Low (can wait a week)

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[X] Feature
[ ] Other... Please describe:
```

### DIFF
--- a/examples/components/CCFRAUD_VERTICAL/traininsilo/aml_comm.py
+++ b/examples/components/CCFRAUD_VERTICAL/traininsilo/aml_comm.py
@@ -285,7 +285,7 @@ class AMLCommSocket(AMLComm):
                 time.sleep(1)
                 conn.setblocking(1)
                 retries += 1
-                self._stats["recv_retries"] += 1
+                self._stats["recv_retries_cnt"] += 1
                 # Send information about failure
                 conn.sendall(pickle.dumps({"flag": FLAGS.FAIL, "data": None}))
                 continue
@@ -593,7 +593,7 @@ class AMLCommRedis(AMLComm):
                 logger.exception(e)
 
                 retries += 1
-                self._stats["recv_retries"] += 1
+                self._stats["recv_retries_cnt"] += 1
                 if retries >= 3:
                     raise e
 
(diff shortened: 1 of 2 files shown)


examples/components/CCFRAUD_VERTICAL/traininsilo/aml_comm.py
```diff
@@ -316,32 +304,49 @@ def send(self, data, destination):
             data: data to be sent
             destination: rank of the receiver node
         """
+        assert destination != self._rank
```
jfomhover, line 307: can you just add some error message here to explain why that's happening? (it won't, but just in case)

examples/components/CCFRAUD_VERTICAL/traininsilo/aml_comm.py
```diff
@@ -316,32 +304,49 @@ def send(self, data, destination):
...
+        self._stats["send_cnt"] += 1
+        self._stats["send_time"] += time.time() - time_start
 
     def recv(self, source):
         """Receives data from the source rank node
 
         Args:
             source: rank of the sender node
         """
+        assert source != self._rank
```
jfomhover, line 332: same comment here

jfomhover (commented):
A couple of minor changes needed, but otherwise looks good

thomasp-ms (approved):
Looks good overall. 
- There are a couple of magic numbers in these files (3, 1024) but they were not introduced in this PR so don't block on that.
- Would it be worth counting the retries as well? (For instance, if the sending times get much higher all of a sudden, is it just because of network latency, or is it because there are direct failures?)

jfomhover (dismissed):

examples/components/CCFRAUD_VERTICAL/traininsilo/aml_comm.py
```diff
@@ -233,22 +234,16 @@ def _send(self, data, destination, flag, ok_flag):
                     )
             except Exception as e:
                 logger.exception(e)
-                tries += 1
+                retries += 1
+                self._stats["send_retries"] += 1
```
thomasp-ms, line 238: Shouldn't this be `send_retries_cnt`? (here and througout, also for recv, also in other file)

jfomhover (approved):

thomasp-ms (changes requested):
There is still a mismatch with `recv_retries_cnt`

PR #312, opened 2023-03-22
User friendly sandboxes

jfomhover:
## Purpose

Our sandboxes are very generic with lots of parameters that most users might not need. Instead, we're providing 7 user friendly sandboxes with preset parameters aligning with different expected behaviors (eyeson, eyesoff, confidential, cpu/gpu, etc).

The old open sandbox is now called "minimal" because it can execute only horizontal FL workloads, but is super quick to setup.

Urgency:
- [ ] Low (can wait a week)

More or less, the old open sandbox is retired.


mlops/bicep/sandbox_fl_eyesoff_gpu.bicep
```diff
@@ -0,0 +1,50 @@
...
+  'australiaeast'
+  'eastus'
+  'westeurope'
+]
+
+@description('The VM used for creating compute clusters in orchestrator and silos.')
+param computeSKU string = 'Standard_NC6'
+
+@description('Uses public network access for the orchestrator storage, allowing it to be eyes-on.')
+param orchestratorEyesOn bool = true
```
garg-amit, line 25: should we keep it `false` by default? (in all eyes-off scripts)

mlops/bicep/sandbox_fl_confidential.bicep
```diff
@@ -0,0 +1,53 @@
...
+
+// please specify the base name for all resources
+@description('Base name of the demo, used for creating all resources as prefix')
+param demoBaseName string = 'fldemo'
+
+@description('Region of the orchestrator (workspace, central storage and compute).')
+param orchestratorRegion string = 'eastus'
+
+@description('List of each region in which to create an internal silo.')
+param siloRegions array = [
```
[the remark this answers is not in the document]
jfomhover: confidential VMs are hard to provision in some regions, eastus worked every time, I used this by default while I figure out a better strategy :-/

mlops/bicep/sandbox_fl_eyesoff_gpu.bicep
```diff
@@ -0,0 +1,50 @@
+// This BICEP script will fully provision a federated learning sandbox
+// with eyes-off orchestrator and silos storages
+// and only one compute (cpu by default)
```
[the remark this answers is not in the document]
jfomhover: oh thx!

mlops/bicep/sandbox_fl_eyesoff_gpu.bicep
jfomhover: that's a good question, I initially did set false by default, then I thought maybe the users would like to actually see just the orchestrator.

mlops/bicep/sandbox_fl_confidential.bicep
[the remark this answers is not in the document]
garg-amit: Oh, I see! You may need to make the corresponding changes in the `sandboxes.md` file then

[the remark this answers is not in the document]
jfomhover: I put a note there already, it's not very publicly displayed, but it's just below the parameter table

"Note: in current sandbox, we're provisioning only in the eastus region by default, to allow for capacity and quick deployment."

mlops/bicep/sandbox_fl_eyesoff_gpu.bicep
jfomhover: I switched to false by default.

garg-amit (dismissed):

mlops/bicep/sandbox_fl_confidential.bicep
```diff
@@ -0,0 +1,49 @@
+// This BICEP script will fully provision a federated learning sandbox
```
thomasp-ms, line 1: This is the exact same comment as the one in `sandbox_fl_eyesoff_cpu.bicep`. Maybe mention that this one is using CC SKU's by default? (and other differences if applicable)

thomasp-ms (changes requested):

mlops/bicep/sandbox_fl_confidential.bicep
jfomhover: Thank you! fixed

thomasp-ms (approved):

PR #317, opened 2023-03-28
Unit tests for shared files in components

majercakdavid:
## Purpose

Test shared files across the components. Coverage:

```
Name                                            Stmts   Miss  Cover   Missing
-----------------------------------------------------------------------------
examples\components\shared\aml_comm.py            257     39    85%   228, 233-235, 268-275, 278, 317-333, 336, 450, 459, 542-559, 564
examples\components\shared\aml_smpc.py             48      0   100%
examples\components\shared\confidential_io.py     102    102     0%   1-267
examples\components\shared\samplers.py             26      0   100%
-----------------------------------------------------------------------------
TOTAL                                             433    141    67%
```

Urgency:
- [ ] Low (can wait a week)

```
[ ] Yes
[ ] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[X] Other... Please describe: Tests
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
pip install coverage
coverage run --source=examples\components\shared -m unittest discover -s tests\examples\components\shared
coverage combine
coverage report -m
```


tests/examples/components/shared/test_aml_comm.py
```diff
@@ -0,0 +1,181 @@
...
+                            use_redis,
+                        ),
+                    )
+                    p1.start()
+                    p2.start()
+                    p1.join()
+                    p2.join()
+
+                    for i in range(2):
+                        self.assertTrue(shared_dict[i])
```
thomasp-ms, line 136: Maybe log the test results explicitly?

tests/examples/components/shared/test_aml_comm.py
```diff
@@ -0,0 +1,181 @@
...
+        while len(self.buffer[session_id]) == 0:
+            time.sleep(1)
+        item = self.buffer[session_id][0]
+        self.buffer[session_id] = self.buffer[session_id][1:]
+        return item
+
+
+class TestAMLComm(unittest.TestCase):
+    def test_aml_comm_socket_simple(self):
+        TEST_MSG = lambda recv: f"Message to {recv}"
```
thomasp-ms, line 91: Would it be worth to also test the case of degenerate empty messages? (I guess we wouldn't encounter them in practice, but just in case, it might be worth testing those edge cases too)

tests/examples/components/shared/test_aml_smpc.py
```diff
@@ -0,0 +1,115 @@
...
+        smpc1.add_remote_public_key(2, smpc2.get_public_key())
+        smpc2.add_remote_public_key(1, smpc1.get_public_key())
+
+        msg_2 = smpc1.encrypt(str.encode(TEST_MSG(2)), 2)
+        msg_1 = smpc2.encrypt(str.encode(TEST_MSG(1)), 1)
+
+        assert smpc1.decrypt(msg_1) == str.encode(TEST_MSG(1))
+        assert smpc2.decrypt(msg_2) == str.encode(TEST_MSG(2))
+
+    def test_aml_smpc_encodings_formats(self):
```
thomasp-ms, line 29: Same comment as in test_aml_comm.py: is it possible to parameterize to avoid code duplication? (and explicitly log what kind of test is being performed, and the reuslts)

tests/examples/components/shared/test_aml_smpc.py
```diff
@@ -0,0 +1,115 @@
...
+
+from cryptography.hazmat.primitives import serialization as crypto_serialization
+
+sys.path.append(
+    os.path.abspath(os.path.join(os.path.dirname(__file__), "..", "..", "..", ".."))
+)
+from examples.components.shared.aml_smpc import AMLSMPC
+
+
+class TestAMLSMPC(unittest.TestCase):
```
thomasp-ms, line 13: Not testing big (or empty) messages in this one?

tests/examples/components/shared/test_samplers.py
```diff
@@ -0,0 +1,147 @@
...
+        # Check that the batches are distributed correctly
+        for i, (batch1, batch2) in enumerate(zip(shared_dict[0][0], shared_dict[1][0])):
+            assert torch.all(batch1 == batch2) and torch.all(
+                batch1
+                == torch.tensor(
+                    range(i * batch_size, min((i + 1) * batch_size, len(data)))
+                )
+            )
+
+    def test_vertically_distributed_batch_sampler_shuffle(self):
```
thomasp-ms, line 94: here too, is it possible to limit code duplication?

thomasp-ms:
Sorry @majercakdavid but I forgot one important comment before I posted the review:

please do include instructions on how to run the tests manually. Those could go in CONTRIBUTING.md for instance

.github/workflows/build-test.yaml
```diff
@@ -24,7 +24,10 @@ jobs:
       if: github.base_ref
 
     - name: Intall python dependencies
-      run: pip install -r examples/pipelines/requirements.txt
+      run: pip install -r examples/pipelines/requirements_test.txt
```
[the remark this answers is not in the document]
majercakdavid: @thomasp-ms @garg-amit the test requirements should be superset of regular requirements. I have separated the file and let pipeline to install both at times where needed. Hope this solves the issue.

tests/examples/components/shared/test_aml_comm.py
```diff
@@ -0,0 +1,181 @@
...
+
+    if rank != 0:
+        # Make sure the first process is started before the others
+        time.sleep(1)
+    if use_redis:
+        comm = TestAMLCommRedis(
+            shared_dict, rank, world_size, run_id, host, port, encryption=encryption
+        )
+    else:
+        comm = AMLCommSocket(
```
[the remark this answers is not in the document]
majercakdavid: both are being tested

tests/examples/components/shared/test_aml_comm.py
majercakdavid: What would be reason for that? Afaik. this is regular assertion you make in unittest.

tests/examples/components/shared/test_aml_smpc.py
majercakdavid: I increased verbosity of an output - what do you mean by code duplication?

tests/examples/components/shared/test_aml_smpc.py
majercakdavid: I can add empty, for large nope, there is packet handling in AMLComm for large messages, there is no special treatment for them in this file.

tests/examples/components/shared/test_aml_comm.py
thomasp-ms: I'm not familiar with the unittest framework. I suggested that because I'm just not sure of what exactly is surfaced when a subtest fails - does it clearly expose the value of the `encrypted` and `use_redis` parameters? If so, then disregard. If not, then knowing exactly which subtest failed would be useful.

tests/examples/components/shared/test_aml_smpc.py
thomasp-ms: Right now, the code in your `test_aml_smpc_basic()` function is exactly the same as the code in the subtest of your `test_aml_smpc_encodings_formats()` function (minus the encoding and format parameters of course). The code for `test_aml_smpc_empty()` is the same too (minus the message generation of course). Why not do it all in one single function using subtests?

I wouldn't worry about the last 2 test functions on overwrite though. They could probably be combined into a single function, but I guess that would require some ugly if statements and make the code harder to read

majercakdavid:
@thomasp-ms I made tests shorter for smpc, however, I would prefer having more explicit tests for aml_sockets. The reason for this is that:

- You can easily deduct already from the name of the test where an error might have happened and during debugging run only this single test to make things right (the amlcomm tests take a bit longer due to sockets)
- You can keep things simple - with a helper function as an abstraction already in place this would make the code even more complex (given you would introduce 3 nested for loops)

.github/actions/run-shared-unit-tests/requirements.txt
```diff
@@ -0,0 +1,7 @@
+--extra-index-url https://download.pytorch.org/whl/cpu
```
garg-amit, line 1: It is probably good to move this file to the `tests` directory and do the respective changes in the `CONTRIBUTING.md` file.

examples/components/shared/aml_comm.py
```diff
@@ -0,0 +1,722 @@
+##########################################################################################
```
garg-amit, line 1: Let’s add an e2e test in the CI/CD that triggers at least [one example pipeline](https://github.com/Azure-Samples/azure-ml-federated-learning/blob/main/.github/workflows/pipeline-e2e-test.yaml#L457) using the shared folder components when any change is detected in the shared folder.

majercakdavid: @garg-amit what's the reasoning behind? The files from shared forlder are never used as of now and whether it is run in CI/CD pipeline in any of the examples is purely dependent on the default config file. Therefore, if any of the files changes the corresponding e2e pipeline is triggered.

thomasp-ms:
> @thomasp-ms I made tests shorter for smpc, however, I would prefer having more explicit tests for aml_sockets. The reason for this is that:
> 
> * You can easily deduct already from the name of the test where an error might have happened and during debugging run only this single test to make things right (the amlcomm tests take a bit longer due to sockets)
> * You can keep things simple - with a helper function as an abstraction already in place this would make the code even more complex (given you would introduce 3 nested for loops)

@majercakdavid Adding a 3rd loop is not that complex, and 3 loops without any code duplication is better than 2 loops with code duplication.

Regarding your first point, isn't it possible to run only one subtest with unittest? I know it is possible with pytest for instance

tests/examples/components/shared/test_aml_comm.py
```diff
@@ -0,0 +1,178 @@
...
+
+            item = self.buffer[session_id][0]
+            self.buffer[session_id] = self.buffer[session_id][1:]
+        finally:
+            self.lock.release()
+        return item
+
+
+class TestAMLComm(unittest.TestCase):
+    def test_aml_comm_full(self):
```
thomasp-ms, line 113: much better!

garg-amit (approved):

